PR #2, opened 2016-07-24
Handle incoming pmtct registrations

gsvr:


gsvr:
@imsickofmaps Please review.

ndoh_hub/utils.py
```diff
@@ -118,18 +118,16 @@ def deactivate_subscription(subscription):
     return patch_subscription(subscription, {"active": False})
 
 
-def get_messageset_short_name(recipient, stage, authority):
-    # Examples:
-    # prebirth.mother.hw_full
-    # prebirth.household.patient
+def get_messageset_short_name(reg_type, authority, weeks):
+    set_number = 1  # default set_number
```
imsickofmaps: can we call the `batch` ... because set has another meaning in this context and using integers it might get complicated.

ndoh_hub/utils.py
```diff
@@ -138,29 +136,11 @@ def get_messageset_schedule_sequence(short_name, weeks):
...
-    if 'postbirth' in short_name:
-        next_sequence_number = 1  # start baby messages at 1
-    elif 'loss' in short_name:
-        next_sequence_number = 1  # always start loss messages at 1
-    else:
-        next_sequence_number = msgs_per_week * (
-            weeks - settings.PREBIRTH_MIN_WEEKS)
-        if next_sequence_number == 0:
-            next_sequence_number = 1  # next_sequence_number cannot be 0
+    next_sequence_number = 1  # to be expanded with momconnect migration
```
imsickofmaps: please double check the scope to see what the stage injection decision was

imsickofmaps:
minor naming change. If the `next_sequence_number` does need to be set, you can open another issue to resolve that. 👍

ndoh_hub/utils.py
gsvr: Scope confusion has ensued. Waiting on a response from matthew and will open another issue depending on what he says.

PR #12, opened 2016-08-19
Accept nurseconnect registrations

gsvr:


gsvr:
@imsickofmaps Please review.

registrations/tasks.py
```diff
@@ -119,6 +124,28 @@ def check_operator_id(self, data_fields, registration):
         else:
             return []
 
+    def check_msisdn_registrant(self, data_fields, registration):
+        if "msisdn_registrant" not in data_fields:
+            return ["Msisdn of Registrant missing"]
```
imsickofmaps: `Msisdn` should be `MSISDN`

registrations/tasks.py
```diff
@@ -198,7 +231,7 @@ def create_subscriptionrequests(self, registration):
             "identity": registration.registrant_id,
             "messageset": msgset_id,
             "next_sequence_number": next_sequence_number,
-            "lang": registration.data["language"],
+            "lang": "eng_ZA",  # NurseConnect is currently only in english
```
imsickofmaps: Not sure this is something worth putting in code in here. The only reason for this is lack of translated content. I think we should move this constraint to the jsbox app only.

imsickofmaps:
Minor changes then 👍  when travis happy

PR #24, opened 2016-08-30
Adjust opt-out strategy

gsvr:
When you use SMS (Stop):
- [x] Set number to opted out
- [x] Deactivate all subscriptions for the identity

When you use USSD:
- [x] Do not set number to opted out
- [x] Nurseconnect optout deactivates nurseconnect subscription only
- [x] Momconnect and pmtct optouts deactivate pmtct and momconnect subscriptions, not nurseconnect

Also does an amazing thing for baby switch that caters for current and migrated pmtct app versions.


imsickofmaps:
👍

PR #28, opened 2016-09-22
Filter subscriptions on identity correctly

gsvr:


gsvr:
@imsickofmaps Please review.

PR #29, opened 2016-09-28
Move to only seed stack system

gsvr:
This should move the hub forward to handling pmtct related tasks when hybrid vumi-seed system is no longer being used.


gsvr:
@imsickofmaps @jpaidoussi Please review. I put a TODO in the code that could be necessary to implement if an update needs to be made to Production code before the full seed stack implementation launches.  The alternative would be to roll back the commits in this branch, make the changes necessary for production, and then re-implement.
There may be a better way around this, but it may also be unnecessary.

PR #33, opened 2016-10-27
Push registrations to Jembi

smn:
Not ready for review yet.

### DIFF
--- a/registrations/tasks.py
+++ b/registrations/tasks.py
@@ -9,7 +9,6 @@ from seed_services_client.identity_store import IdentityStoreApiClient
 from seed_services_client.service_rating import ServiceRatingApiClient
 
 from ndoh_hub import utils
-from .models import Registration, SubscriptionRequest
 
 
 is_client = IdentityStoreApiClient(
@@ -302,6 +301,7 @@ class ValidateSubscribe(Task):
             "schedule": msgset_schedule
         }
         self.l.info("Creating SubscriptionRequest object")
+        from .models import SubscriptionRequest
         SubscriptionRequest.objects.create(**subscription)
         self.l.info("SubscriptionRequest created")
 
@@ -351,6 +351,7 @@ class ValidateSubscribe(Task):
         """
         self.l = self.get_logger(**kwargs)
         self.l.info("Looking up the registration")
+        from .models import Registration
         registration = Registration.objects.get(id=registration_id)
 
         reg_validates = self.validate(registration)
@@ -365,9 +366,6 @@ class ValidateSubscribe(Task):
             if "pmtct" in registration.reg_type:
                 self.set_risk_status(registration)
 
-            push_registration_to_jembi.apply_async(kwargs={
-                'registration_id': str(registration.id)
-            })
             self.l.info("Scheduling task push_registration_to_jembi")
 
             self.l.info("Task executed successfully")
--- a/registrations/tasks.py
+++ b/registrations/tasks.py
@@ -399,18 +399,14 @@ class PushRegistrationToJembi(Task):
 
     def get_subscription_type(self, authority):
         authority_map = {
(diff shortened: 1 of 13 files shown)


registrations/tasks.py
```diff
@@ -375,6 +377,79 @@ def run(self, registration_id, **kwargs):
...
+        }
+
+        # Self registrations on all lines should use cmsisdn as dmsisdn too
+        if registration.hcw_msisdn is None:
+            json_template["dmsisdn"] = registration.mom_msisdn
+
+        if registration.authority == 'clinic':
+            json_template["edd"] = registration.mom_edd.strftime("%Y%m%d")
+
+        return json_template
```
smn: @jpaidoussi could I get your input on this? This is obviously failing because I'm trying to repopulate a JSON payload for an API request using a new style registration and the fields either don't exist or aren't of the same name. Could you help me with how these should be mapped?

smn: I'm guessing I'm going to have to talk to the identity store to try and get some of these bits of information?

jpaidoussi: Most of the fields will be mapped to keys in the data json field in registrations. 

Here I did the mapping for the migration:
https://github.com/praekeltfoundation/momconnect-migration-tools/blob/develop/migrationtools/databases.py#L197

smn:
@jpaidoussi ready for review

registrations/tasks.py
```diff
@@ -375,6 +382,118 @@ def run(self, registration_id, **kwargs):
...
+            result.raise_for_status()
+            return result.text
+        except (HTTPError,) as e:
+            # retry message sending if in 500 range (3 default retries)
+            if 500 < e.response.status_code < 599:
+                raise self.retry(exc=e)
+            else:
+                raise e
+        except (Exception,) as e:
+            self.l.error('Problem posting JSON to Jembi', exc_info=True)
```
jpaidoussi: Perhaps we can include the registration uuid at this point in case we need to rerun this manually.

jpaidoussi:
Looks good, 👍

registrations/tasks.py
```diff
@@ -375,6 +382,118 @@ def run(self, registration_id, **kwargs):
...
+        NOTE:   this is a convenience method to map the new "source"
+                back to ndoh-control's "authority" fields to maintain
+                backwards compatibility with existing APIs
+        """
+        return {
+            'CLINIC USSD app': 'clinic',
+            'PUBLIC USSD app': 'personal',
+            'CHW USSD app': 'chw',
+            'NURSE USSD App': 'nurseconnect',
+        }.get(source.name, 'unknown')  # See note at get_subscription_type
```
smn: @jpaidoussi I'm unsure about this. The `Source` model has an `authority` attribute, should I be using that? This mapping only applies to old "migrated" records, not for new registrations I think.

smn: I've changed this to now use the source's authority instead.

smn:
@jpaidoussi could you review my last commit please?

registrations/tasks.py
```diff
@@ -420,22 +424,9 @@ def get_dob(self, mom_dob):
...
-            'CLINIC USSD app': 'clinic',
-            'PUBLIC USSD app': 'personal',
-            'CHW USSD app': 'chw',
-            'NURSE USSD App': 'nurseconnect',
-        }.get(source.name, 'unknown')  # See note at get_subscription_type
-
     def build_jembi_json(self, registration):
         """ Compile json to be sent to Jembi. """
-        authority = self.get_authority_from_source(registration.source)
+        authority = registration.source.authority
```
jpaidoussi: The registration source model does have an authority field but it contains data like "hw_limited" or "hw_full" or "patient". It doesn't directly map to the system (line / app) as different systems can have the same authority level.

smn: ok reverting.

PR #38, opened 2016-11-14
Add nurse registrations

rudigiesler:


registrations/tasks.py
```diff
@@ -347,6 +347,23 @@ def set_risk_status(self, registration):
...
+            registration.source)
+        task = {
+            'nurse': push_nurse_registration_to_jembi,
+            'personal': push_registration_to_jembi,
+            'optout': push_registration_to_jembi,
+            'clinic': push_registration_to_jembi,
+            'chw': push_registration_to_jembi,
+        }.get(authority, push_registration_to_jembi)
+
+        return task.delay(registration_id=str(registration.pk))
```
smn: I'm not entirely sure what the difference is between `task.delay` and `task.apply_async`. I noticed we're using `apply_async` below.

registrations/tasks.py
```diff
@@ -435,8 +440,66 @@ def get_authority_from_source(self, source):
...
+
+
+class PushRegistrationToJembi(BasePushRegistrationToJembi, Task):
+    """ Task to push registration data to Jembi
+    """
+    name = "ndoh_hub.registrations.tasks.push_registration_to_jembi"
+    l = get_task_logger(__name__)
+    URL = "%s/subscription" % settings.JEMBI_BASE_URL
+
+    def get_subscription_type(self, authority):
```
smn: Where is function now called from?

rudigiesler: I think the diff is a little strange, it looks like I renamed `PushRegistrationToJembi` to `BasePushRegistrationToJembi` and moved functions out to a new `PushRegistrationToJembi` class; when instead I made a new `BasePushRegistrationToJembi` and moved some functions out from `PushRegistrationToJembi` to it, so you kind of have to read the diff in reverse.

This function is still in the `PushRegistrationToJembi` class, and called from the `build_jembi_json` function.

registrations/tasks.py
smn: :+1:

smn:
@rudigiesler :+1: from me and I'm going to merge this to test it in production.

PR #39, opened 2016-11-15
Feature/3rd party registrations

jpaidoussi:


registrations/views.py
```diff
@@ -137,3 +156,109 @@ def get(self, request, *args, **kwargs):
...
+                'msisdn_device': device,
+                'id_type': id_type,
+                'language': lang_code,
+                'edd': serializer.validated_data['mom_edd'],
+                'faccode': serializer.validated_data['clinic_code'],
+                'consent': serializer.validated_data['consent'],
+
+            }
+            reg = Registration.objects.create(
+                reg_type='momconnect_prebirth',
```
jpaidoussi: I am not sure if this is the only type of 3rd party registration we are supporting?

registrations/views.py
```diff
@@ -137,3 +156,109 @@ def get(self, request, *args, **kwargs):
...
+        if serializer.is_valid():
+            mom_msisdn = serializer.validated_data['mom_msisdn']
+            hcw_msisdn = serializer.validated_data['hcw_msisdn']
+            lang_code = serializer.validated_data['mom_lang']
+            lang_code = transform_language_code(lang_code)
+            if mom_msisdn != hcw_msisdn:
+                # Get or create HCW Identity
+                result = is_client.get_identity_by_address(
+                    'msisdn', hcw_msisdn)
+                if 'results' in result and not result['results']:
```
gsvr: Some clarification would be helpful here, this line is hard to read.

registrations/views.py
```diff
@@ -137,3 +156,109 @@ def get(self, request, *args, **kwargs):
...
+                    },
+                    'operator_id': operator,
+                    'lang_code': lang_code,
+                    'id_type': id_type,
+                    'mom_dob': serializer.validated_data['mom_dob'],
+                    'edd': serializer.validated_data['mom_edd'],
+                    'faccode': serializer.validated_data['clinic_code'],
+                    'consent': serializer.validated_data['consent'],
+                    'mha': serializer.validated_data['mha'],
+                    'swt': serializer.validated_data['swt'],
```
gsvr: These fields (`operator_id` to `swt`) should be inside the `details` object.

registrations/views.py
```diff
@@ -137,3 +156,109 @@ def get(self, request, *args, **kwargs):
...
+                        'addresses': {
+                            'msisdn': {
+                                mom_msisdn: {'default': True}
+                            }
+                        }
+                    },
+                    'operator_id': operator,
+                    'lang_code': lang_code,
+                    'id_type': id_type,
+                    'mom_dob': serializer.validated_data['mom_dob'],
```
gsvr: It looks like we're forcing the external api calls to provide a `mom_dob`, which is fine. Just note that in the JS apps we don't always ask for the mom_dob, so for clinic we may extract the dob from the id number and we may not have it for the passport flow.

registrations/views.py
```diff
@@ -137,3 +156,109 @@ def get(self, request, *args, **kwargs):
...
+                    'edd': serializer.validated_data['mom_edd'],
+                    'faccode': serializer.validated_data['clinic_code'],
+                    'consent': serializer.validated_data['consent'],
+                    'mha': serializer.validated_data['mha'],
```
gsvr: I would probably pass `mha` and `swt` to the registration rather than save it in the identity.

registrations/views.py
```diff
@@ -137,3 +156,109 @@ def get(self, request, *args, **kwargs):
...
+            reg_data = {
+                'operator_id': operator,
+                'msisdn_registrant': mom_msisdn,
+                'msisdn_device': device,
+                'id_type': id_type,
+                'language': lang_code,
+                'edd': serializer.validated_data['mom_edd'],
+                'faccode': serializer.validated_data['clinic_code'],
+                'consent': serializer.validated_data['consent'],
+
```
gsvr: if id_type == "passport", add fields passport_no, passport_origin
if id_type == "sa_id", add fields sa_id_no, mom_dob
if id_type == "none", add field mom_dob

registrations/views.py
```diff
@@ -137,3 +156,109 @@ def get(self, request, *args, **kwargs):
...
+                            'msisdn': {
+                                mom_msisdn: {'default': True}
+                            }
+                        }
+                    },
+                    'operator_id': operator,
+                    'lang_code': lang_code,
+                    'id_type': id_type,
+                    'mom_dob': serializer.validated_data['mom_dob'],
+                    'edd': serializer.validated_data['mom_edd'],
```
gsvr: In the JS apps we've called this `last_edd` on the identity, since it will get overwritten if the mother registers for another pregnancy.

registrations/views.py
```diff
@@ -137,3 +156,109 @@ def get(self, request, *args, **kwargs):
...
+                    identity['details']['sa_id_no'] = (
+                        serializer.validated_data['mom_id_no'])
+                elif id_type == 'passport':
+                    identity['details']['passport_origin'] = (
+                        serializer.validated_data['mom_passport_origin'])
+                    identity['details']['passport_no'] = (
+                        serializer.validated_data['mom_id_no'])
+                mom_identity = is_client.create_identity(identity)
+            else:
+                mom_identity = result['results'][0]
```
gsvr: I'm not sure if you've discussed with the strategy team what should happen if an external registration tries to register someone with a number that we already have an identity for, it looks like you're okay with just accepting it. However, I do think you should then be updating the identity details here.

registrations/views.py
gsvr: We set two other fields in the js apps: `last_mc_reg_on` and `source`.
`source` should keep track of which app created the identity, in this case `"external"` is probably good enough, otherwise you could maybe use the `user` or something. (This is not critical info, though)
`last_mc_reg_on` I would set to `serializer.validated_data['authority']`. This is used for some routing in `ussd_public.js`

registrations/views.py
gsvr: `reg_data` will also depend on the `authority` in `serializer.validated_data`.  To map to the new system registrations:
clinic = hw_full
chw = hw_partial
personal/public = patient

smn:
:+1: from me on this PR.

@gsvr any further feedback from you?

gsvr:
:+1:

PR #43, opened 2016-11-16
Added source lookup with authority mapping

jpaidoussi:


registrations/views.py
```diff
@@ -166,14 +166,22 @@ def post(self, request):
...
+            authority = serializer.validated_data['authority']
+            # load the users sources with authority mapping
+            if authority == 'chw':
+                source_auth = 'hw_limited'
+            elif authority == 'clinic':
+                source_auth = 'hw_full'
+            else:
+                source_auth = 'patient'
+            source = Source.objects.get(user=self.request.user,
+                                        authority=source_auth)
```
jpaidoussi: @gsvr This lookup will support a single user with multiple sources. Are there any implications on other parts of the code though? I think as long as the source and user used for external registrations is never used elsewhere is should be fine right?

registrations/tests.py
```diff
@@ -406,6 +406,22 @@ def make_source_normaluser(self):
         }
         return Source.objects.create(**data)
 
+    def make_external_source_limited(self):
+        data = {
+            "name": "test_source_external_limited",
+            "authority": "hw_limited",
```
gsvr: hw_partial

registrations/views.py
```diff
@@ -166,14 +166,22 @@ def post(self, request):
...
         serializer = ThirdPartyRegistrationSerializer(data=request.data)
         if serializer.is_valid():
             mom_msisdn = serializer.validated_data['mom_msisdn']
             hcw_msisdn = serializer.validated_data['hcw_msisdn']
             lang_code = serializer.validated_data['mom_lang']
             lang_code = transform_language_code(lang_code)
+            authority = serializer.validated_data['authority']
+            # load the users sources with authority mapping
+            if authority == 'chw':
+                source_auth = 'hw_limited'
```
gsvr: hw_partial

registrations/views.py
gsvr: I can't foresee any problems.

registrations/views.py
jpaidoussi: I got limited from here: https://github.com/praekeltfoundation/ndoh-hub/blob/feature%2Fextend-source-lookup/registrations/models.py#L19 which is right?

gsvr: my mistake, hw_limited is correct

PR #37, opened 2016-11-16
Reject registrations with invalid JSONField details

gsvr:


gsvr:
@smn @jpaidoussi Please review.

registrations/models.py
```diff
@@ -74,6 +76,202 @@ class Registration(models.Model):
...
+            self._check_field_values(required_fields)
+
+        # Raise the validation errors encountered if any
+        if len(self.v) > 0:
+            raise ValidationError(self.v)
+        else:
+            self.validated = True
+
+    def save(self, *args, **kwargs):
+        self.clean()
```
jpaidoussi: I think full_clean() should be called here instead of just clean()?

registrations/serializers.py
```diff
@@ -39,6 +39,11 @@ class Meta:
                   'source', 'created_at', 'updated_at', 'created_by',
                   'updated_by')
 
+    def validate(self, attrs):
+        instance = Registration(**attrs)
+        instance.clean()
```
jpaidoussi: I am not sure if DRF's ModelSerializer will directly call full_clean on the model instance but if not we should call it here instead of clean.

nathanbegbie:
@rudigiesler should we close this PR?

rudigiesler:
@nathanbegbie I mean, we should probably do something similar to this as cleanup on this code base, but yes, it can probably be closed. I think it's too old to be of much use, better to do it from scratch with the more up-to-date code base.

PR #54, opened 2016-11-17
ensure change_pk sent to celery is a string

miltontony:


miltontony:
@rudigiesler review please

PR #58, opened 2016-11-17
Handle case when operator_id is not available

miltontony:


miltontony:
ready for review

smn:
:+1: when travis and pierre :)

smn:
:+1: when travis

PR #60, opened 2016-11-18
op is expected to be a string

miltontony:


miltontony:
@rudigiesler review please

PR #62, opened 2016-11-18
dhis2 expects inbound and outbound timestamps. We were only sending outbound

miltontony:


miltontony:
@rudigiesler review please

PR #68, opened 2016-11-25
set faccode to None when not available

miltontony:


miltontony:
@rudigiesler review please

PR #73, opened 2016-12-07
Notify jembi of baby loss change

erikh360:
This adds functionality that will send a payload to jembi when a baby loss change is created.

Also, a management command to resubmit any missed changes

### DIFF
--- a/changes/tasks.py
+++ b/changes/tasks.py
@@ -251,6 +251,7 @@ class ValidateImplement(Task):
 
         if switched is True:
             self.l.info("Completed PMTCT switch to loss")
+            push_momconnect_babyloss_to_jembi.delay(str(change.pk))
             return "Completed PMTCT switch to loss"
         else:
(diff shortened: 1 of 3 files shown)


changes/tasks.py
```diff
@@ -251,6 +251,7 @@ def pmtct_loss_switch(self, change):
 
         if switched is True:
             self.l.info("Completed PMTCT switch to loss")
+            push_momconnect_babyloss_to_jembi.delay(str(change.pk))
```
KaitCrawford: I think this needs to be added to loss switches from other services as well. So here https://github.com/praekeltfoundation/ndoh-hub/blob/32f384bf43a4089e953be84a8e4c8338ebcc93da/changes/tasks.py#L314

rudigiesler:
@KaitCrawford Yes I agree, we will need to have a way to send the historical data to Jembi.

KaitCrawford:
:+1: thanks for taking this on :cake:

PR #74, opened 2016-12-15
Support for encdate to 3rd party registrations

erikh360:
This PR adds the optional field "encdate" to the 3rd party integrations, if it is not included in the payload the old functionality will apply(get_timestamp).

The reason for this change is to make the "encdate" field more accurate when sending historical registrations to the hub.


smn:
thanks for doing this, is there a way to test that encdate is being set properly on the registration created?

erikh360:
@smn I have added a check to see if the encdate is set properly on the registration

smn:
thanks! :+1: :cake:

PR #75, opened 2016-12-15
Notify jembi of baby switch change

erikh360:
This adds functionality that will send a payload to jembi when a baby switch change is created.

Also, a management command to resubmit any missed changes


erikh360:
@KaitCrawford please review.

KaitCrawford:
:+1:

PR #78, opened 2016-12-22
management command to generate pmtct registrations csv

smn:


smn:
@jpaidoussi @rudigiesler @KaitCrawford @devchima ready for review

rudigiesler:
Looks good. Would prefer Django queryset rather than raw SQL, but looks good to me. :+1:

smn:
@rudigiesler yeah, I considered it but that would tie it to a model even though the query itself doesn't necessary always apply to a model.

PR #81, opened 2017-01-03
Feature/seed 1016 add kpi metrics

jpaidoussi:


registrations/signals.py
```diff
@@ -8,3 +8,14 @@ def psh_validate_subscribe(sender, instance, created, **kwargs):
         from .tasks import validate_subscribe
         validate_subscribe.apply_async(
             kwargs={"registration_id": str(instance.id)})
+
+
+def psh_fire_created_metric(sender, instance, created, **kwargs):
```
erikh360: Can we add a test to make sure this fires when it needs to?

PR #82, opened 2017-01-03
Update pmtct opt out url and management command

erikh360:
The url is different for PMTCT opt outs and I fixed the management command to use the correct class.


erikh360:
@KaitCrawford please review, thanks.

KaitCrawford:
:+1:

PR #83, opened 2017-01-12
Adds a call to identity store for msisdn

erikh360:
The PMTCT registrations does not have any data we can use to identify a patient, so we do a call to the identity store to get the msisdn. We can then send the registration to Jembi.

### DIFF
--- a/registrations/tasks.py
+++ b/registrations/tasks.py
@@ -409,12 +409,13 @@ class BasePushRegistrationToJembi(object):
             return mom_msisdn.replace('+', '') + '^^^ZAF^TEL'
         elif registrant_id:
             identity = is_client.get_identity(registrant_id)
-            msisdns = \
-                identity['details'].get('addresses', {}).get('msisdn', {})
+            if identity:
+                msisdns = \
+                    identity['details'].get('addresses', {}).get('msisdn', {})
 
-            for key in msisdns:
-                if not msisdns[key].get('optedout', False):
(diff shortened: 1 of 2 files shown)


registrations/tasks.py
```diff
@@ -399,14 +399,22 @@ class BasePushRegistrationToJembi(object):
...
+                       mom_msisdn=None, registrant_id=None):
         if id_type == 'sa_id':
             return id_no + "^^^ZAF^NI"
         elif id_type == 'passport':
             return id_no + '^^^' + passport_origin.upper() + '^PPN'
-        else:
+        elif mom_msisdn:
             return mom_msisdn.replace('+', '') + '^^^ZAF^TEL'
+        elif registrant_id:
+            identity = is_client.get_identity(registrant_id)
```
jpaidoussi: You might need to protect against getting None from get_identity

registrations/tasks.py
```diff
@@ -399,14 +399,23 @@ class BasePushRegistrationToJembi(object):
...
+            if identity:
```
jpaidoussi: As you have no else clause for this if, are all the callers of this function ok with receiving an implicit null in this, possibly rare, case?

erikh360: Yes, it will then fail on Jembi's side with a error we can use, previously it failed because mom_msisdn was None, the errors in the log files weren't helpful and it never got to Jembi. Is this ok?

PR #86, opened 2017-01-17
Command to fix unvalidated PMTCT optouts

erikh360:
Since the migration the new PMTCT registrations have been failing to validate, this is because the `edd` field was not populated.

This commands looks for a related subscription of a different type that is validated to get the `edd` date. It will then resubmit it for validation.

### DIFF
--- a/registrations/management/commands/fix_pmtct_registrations.py
+++ b/registrations/management/commands/fix_pmtct_registrations.py
@@ -1,0 +1,42 @@
+from django.core.management.base import BaseCommand
+
+from registrations.models import Registration
+from registrations.tasks import validate_subscribe
+
+
+class Command(BaseCommand):
+    help = ("Find and fixes all PMTCT registrations that aren't validated"
+            "because of the edd date not being populated")
+
+    def handle(self, *args, **kwargs):
+
+        registrations = Registration.objects.filter(
+            validated=False,
+            reg_type__in=("pmtct_prebirth", "pmtct_postbirth")).iterator()
+
+        updates = 0
+
+        for registration in registrations:
+
+            related_regs = Registration.objects.filter(
+                    validated=True,
+                    registrant_id=registration.registrant_id
+                ).exclude(reg_type__in=("pmtct_prebirth", "pmtct_postbirth"))
+
+            for related_reg in related_regs:
+
+                if related_reg.data.get("edd"):
+                    del registration.data["invalid_fields"]
+                    registration.data["edd"] = related_reg.data["edd"]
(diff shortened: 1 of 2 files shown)


bruskiza:
@KaitCrawford @rudigiesler can u help @erikh360 out here please?
cc @smn

registrations/management/commands/fix_pmtct_registrations.py
```diff
@@ -0,0 +1,43 @@
...
+            related_regs = Registration.objects.filter(
+                    validated=True,
+                    registrant_id=registration.registrant_id
+                ).exclude(reg_type__in=("pmtct_prebirth", "pmtct_postbirth")).\
+                order_by('-created_at')
+
+            for related_reg in related_regs:
+
+                if related_reg.data.get("edd"):
+                    del registration.data["invalid_fields"]
```
rudigiesler: Maybe we should have a check here to make sure that edd is the only missing field (and if it's not, maybe just remove edd from the list). Is this a list that's easily parsable, or is it a human readable string? If it's a string I think it will be too finicky and you can just ignore this suggestion.

rudigiesler: I'm wondering if there's a case where the `invalid_fields` field does not exist, but the registration still doesn't have the edd field. Maybe if the celery task crashed early or something. Maybe we should do a `registration.data.pop('invalid_fields', None)` or wrap it in a try catch so that this doesn't crash if the field doesn't exist.

registrations/management/commands/fix_pmtct_registrations.py
```diff
@@ -0,0 +1,48 @@
...
+                order_by('-created_at')
+
+            updated = False
+            for related_reg in related_regs:
+                for field in ['edd', 'language', 'mom_dob', 'operator_id']:
+
+                    if (related_reg.data.get(field) and
+                            not registration.data.get(field)):
+                        registration.data[field] = related_reg.data[field]
+                        updated = True
```
rudigiesler: Wouldn't this mean that we're using the oldest registration we have for the field value? It also doesn't assure us that all the fields that are missing have been updated.

We could also let the database do this work. `related_regs.filter(data__has_key=field).first()` will return the most recent registration with that field, or `None` if a registration doesn't exist.

So a loop like `for field in set(('edd', ...)).difference(registration.data.keys())`, and in there we have the `related_regs.filter(data__has_key=field).first()`, and if any of the fields result in `None`, then we can not update the registration (since we don't have all the fields we need), and we skip that registration.

PR #88, opened 2017-01-19
Change encdate to create date of item

erikh360:
One of the fields that is being sent to Jembi is `encdate`, this requires the timestamp of when the event happened, because we sometimes send through items late, it is better to use the item create date instead of current timestamp.


erikh360:
Ready for review!

PR #89, opened 2017-01-31
Feature/momza 480 optout inactive msisdns

jpaidoussi:
This management command is expected to be run on a local instance of the ndoh-hub code and communicate with the live services entirely over HTTP.

### DIFF
--- a/tools/add-identity-to-jembi-csv.py
+++ b/tools/add-identity-to-jembi-csv.py
@@ -1,0 +1,43 @@
+import csv
+
+from sqlalchemy import create_engine, MetaData
+from sqlalchemy.sql import select
+
+
+# This is an example script and should not be used directly.
+
+# Takes a Jembi provided CSV file of MSISDNs to deactivate and looks up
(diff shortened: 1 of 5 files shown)


tools/add-identity-to-jembi-csv.py
```diff
@@ -0,0 +1,43 @@
...
+    for row in reader:
+        msisdn = '+{0}'.format(row[1])
+        res = con.execute(
+            select([tbl.c.id]).where(tbl.c.msisdn == msisdn)
+        ).fetchone()
+        new_row = list(row)
+        if res:
+            new_row.append(res['id'])
+        else:
+            print('no msisdn match')
```
smn: Do we want to include the msisdn we're trying to match here? If not, why are we printing this?

jpaidoussi: Good point, it was initially just for me to get an idea of how many weren't matching but it would be more useful with the msisdn

PR #94, opened 2017-02-28
Added filter for validated

erikh360:


registrations/management/commands/report_pmtct_risks.py
```diff
@@ -111,7 +111,8 @@ def add_to_result(risk, reg):
 
         else:
             registrations = Registration.objects.filter(
-                Q(reg_type='pmtct_postbirth') | Q(reg_type='pmtct_prebirth'))
+                Q(reg_type='pmtct_postbirth') | Q(reg_type='pmtct_prebirth'),
+                validated=True)
```
rudigiesler: While we're at it, do we also maybe want to add `.iterator()`?

erikh360: Good idea.

PR #100, opened 2017-06-01
Add change MomConnect MSISDN change type

rudigiesler:
This change type should allow momconnect apps to change the MSISDN of a user.

### DIFF
--- a/changes/tasks.py
+++ b/changes/tasks.py
@@ -398,11 +398,13 @@ class ValidateImplement(Task):
 
     def momconnect_change_msisdn(self, change):
         """
-        MSISDN change should change the default msisdn of the identity.
+        MSISDN change should change the default msisdn of the identity, while
+        storing information on the identity to allow us to see the history
+        of the change.
         """
         self.l.info("Starting MomConnect MSISDN change")
 
-        new_msisdn = change.data['msisdn']
+        new_msisdn = change.data.pop('msisdn')
 
         self.l.info("Fetching identity")
         identity = is_client.get_identity(change.registrant_id)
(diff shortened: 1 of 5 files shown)


KaitCrawford (approved):
I'm adding a check in the jsbox that prevents a user changing to an active msisdn on another identity. Since it'll be in the jsbox it might not be needed here but something to consider.
Either way :+1: looks good to me.

changes/tasks.py
```diff
@@ -524,6 +560,17 @@ def check_momconnect_change_language(self, data_fields, change):
         else:
             return []
 
+    def check_momconnect_change_msisdn(self, data_fields, change):
+        """
+        Ensures that a new msisdn is provided, and that it is a valid msisdn.
+        """
+        if "msisdn" not in data_fields:
```
KaitCrawford: As for a language change, I think the old msisdn should also be in the data. that way it's explicit and more easily traceable.

rudigiesler: The tricky bit here is that it's difficult to know what an identity's old address is, and it could also be multiple addresses if there's no or more than one default set. Should we just be storing the whole dictionary of addresses? Or should we specifically look at which address/addresses and just store those?

Technically, since it doesn't remove any addresses, the old address will still be there. What we can't tell is which ones, if any, were default addresses before we made the change.

KaitCrawford: hmmm good point. It would be nice to keep the knowledge of which number was the default but I'm not sure how to do that. Perhaps we could populate a "previous_default" field?
another thought was we should probably record the msisdn that the change request came from, although I don't think that's specific to the msisdn change.

rudigiesler: The issue with a `previous_default` field is that there could be multiples (there could be multiple defaults, or no default so we have to include all numbers). Then, for the defaults, do we also want to include all the additional data?

I think we should either just store everything under `msisdn`, or we store all the details of everything under `msisdn` that we consider "default" (either default, or everything if there's no default).

KaitCrawford: I'd go with option there.
But I'm now not sure we should do this at all, the main reason being that if the user chooses to delete their data then we will have to go through all of their changes and delete the phone numbers (and id numbers for the other changes we need to implement). :confused:

rudigiesler: That's a good point. What if we put the ID of the change object on the blob of the address details?

KaitCrawford: So that we can remove the numbers from the change object when the identity gets changed? I guess that works.

It isn't an option for other field changes though, like id numbers, since those don't have detail blobs.

rudigiesler: I've made the changes, see if they're okay?

KaitCrawford (approved):
:+1: that works :)

PR #104, opened 2017-06-07
Remove personal information from registrations

rudigiesler:
Currently, there is personal information stored in registrations, so when we "forget" a user in the identity store, wiping the information from there doesn't clear all the information we have on them.

Instead, once we are done with this information, we should make sure it's stored on the identity, in case we need it later, and then we should remove it from the registration object in the database.

### DIFF
--- a/registrations/tasks.py
+++ b/registrations/tasks.py
@@ -402,9 +402,14 @@ def remove_personally_identifiable_fields(registration_id):
     fields = set((
         'id_type', 'mom_dob', 'passport_no', 'passport_origin', 'sa_id_no',
         'language', 'consent')).intersection(registration.data.keys())
-    identity = is_client.get_identity(registration.registrant_id)
-    for field in fields:
-        identity['details'][field] = registration.data.pop(field)
+    if fields:
+        identity = is_client.get_identity(registration.registrant_id)
+
+        for field in fields:
+            identity['details'][field] = registration.data.pop(field)
+
+        is_client.update_identity(
+            identity['id'], {'details': identity['details']})
 
     msisdn_fields = set((
         'msisdn_device', 'msisdn_registrant')
@@ -425,8 +430,6 @@ def remove_personally_identifiable_fields(registration_id):
         field = field.replace('msisdn', 'uuid')
         registration.data[field] = field_identity['id']
 
-    is_client.update_identity(
-        identity['id'], {'details': identity['details']})
     registration.save()
 
 
--- a/registrations/tasks.py
+++ b/registrations/tasks.py
@@ -6,12 +6,14 @@ import requests
 from requests.exceptions import HTTPError
 
 from django.conf import settings
+from celery import chain
 from celery.task import Task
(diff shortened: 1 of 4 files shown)


registrations/tasks.py
```diff
@@ -392,6 +391,48 @@ def run(self, registration_id, **kwargs):
...
+def remove_personally_identifiable_fields(registration_id):
+    """
+    Saves the personally identifiable fields to the identity, and then
+    removes them from the registration object.
+    """
+    registration = Registration.objects.get(id=registration_id)
+
+    fields = set((
+        'id_type', 'mom_dob', 'passport_no', 'passport_origin', 'sa_id_no',
+        'language', 'consent')).intersection(registration.data.keys())
```
KaitCrawford: Is 'language' not something we'd want to keep for aggregated or demographic data? Especially for investigations like the ones we're currently doing

registrations/tasks.py
```diff
@@ -380,7 +373,13 @@ def run(self, registration_id, **kwargs):
                 self.set_risk_status(registration)
 
             self.l.info("Scheduling registration push to Jembi")
-            self.send_to_jembi(registration)
+            jembi_task = BasePushRegistrationToJembi.\
+                get_jembi_task_for_registration(registration)
+            task = chain(
+                jembi_task.si(str(registration.pk)),
+                remove_personally_identifiable_fields.si(str(registration.pk))
```
KaitCrawford: Does this only trigger if the previous task was successful? We might not want to remove the fields if we failed to send the info to Jembi.

KaitCrawford:
I have 2 concerns about landing this PR:
1) I think we will need to do an audit of what assumes this data is on the hub (management commands, ci, jsbox apps, etc)
2) This will be quite QA intensive and since we don't have versioned releases here yet that will further hold up the other work you did for the user data jsbox app.

registrations/tasks.py
rudigiesler: Possibly, but it seems to be a fairly high risk item to keep, looking at our security research, so I opted to take it out.

We can always get that data back by looking up the identity, unless they've opted out, in which case we don't want to use them for the investigation anyway.

registrations/tasks.py
rudigiesler: Yup, that was my understanding of `chain` from the celery docs. http://docs.celeryproject.org/en/latest/userguide/canvas.html#chains

rudigiesler:
@KaitCrawford Yes, that's a "part 2" of this, to have a look how this affects everything else, and to make those things rather look at the identity store.

Noted. I won't land these PRs yet, until the other work has been deployed.

registrations/tasks.py
KaitCrawford: fair enough

rudigiesler:
@KaitCrawford I've switched the base branch from develop to a new branch, so that we can merge all of these commits into there, and then merge that to develop when we're ready.

KaitCrawford:
Good idea :+1:

PR #105, opened 2017-06-08
Update registration management commands for identity information removal

rudigiesler:


registrations/management/commands/report_pmtct_risks.py
```diff
@@ -115,7 +122,7 @@ def add_to_result(risk, reg):
                 validated=True)
 
             for registration in registrations.iterator():
-
+                add_personally_identifiable_fields(registration)
```
KaitCrawford: Shouldn't we call the task to remove this data again when we're done here? Or would that get called somewhere after this already?

rudigiesler: This task just adds it to the current registration object, it doesn't actually save the changes to the database. So as long as we don't save it, the changes will be lost when the object gets cleaned from memory.

registrations/management/commands/report_pmtct_risks.py
```diff
@@ -103,8 +104,14 @@ def add_to_result(risk, reg):
                     validated=True)
 
                 for registration in registrations:
+                    mom_dob = registration["data"].get("mom_dob")
+                    if not mom_dob:
+                        identity = self.get_identity(
+                            ids_client, registration["registrant_id"])
+                        mom_dob = identity["details"].get("mom_dob")
```
rudigiesler: In that case, `.get` returns `None`. I think the code downstream though might have an issue being passed `None`, should I maybe add a check and skip processing the registration if that's the case?

registrations/management/commands/report_pmtct_risks.py
KaitCrawford: hmmmmm ok. that complicates things a little bit if we decide we do want to save it but it should be fine. Maybe add a comment in the task so that it's clear the changes made there aren't saved?

rudigiesler: Yup, can do it. It's a little complicated since it's not a task, but I wanted it in the same file as the remove personal fields task, since they're related.

registrations/management/commands/report_pmtct_risks.py
KaitCrawford: Yeah, we probably don't need to update the mother's risk status if she's opted out.

rudigiesler: Done

registrations/management/commands/report_pmtct_risks.py
rudigiesler: Done

rudigiesler:
@erikh360 Since @KaitCrawford is on leave, would you mind giving the changes a review?

@dubesoftware

PR #106, opened 2017-06-13
New admin optout endpoint

erikh360:


changes/views.py
```diff
@@ -72,3 +78,54 @@ def post(self, request, *args, **kwargs):
...
+        sbm_client = StageBasedMessagingApiClient(
+            api_url=settings.STAGE_BASED_MESSAGING_URL,
+            auth_token=settings.STAGE_BASED_MESSAGING_TOKEN
+        )
+
+        active_subs = sbm_client.get_subscriptions(
+            {'identity': identity_id, 'active': True}
+        )["results"]
+
+        actions = []
```
rudigiesler: We probably want this to be a `set` rather than a list. 

eg. If they have multiple nurseconnect subscriptions, we don't want to create multiple `nurse_optout` change objects, we just want to create one that will remove all of the subscriptions.

changes/views.py
```diff
@@ -72,3 +78,54 @@ def post(self, request, *args, **kwargs):
...
+            data = utils.json_decode(request.body)
+        except ValueError as e:
+            return JsonResponse({'reason': "JSON decode error",
+                                'details': six.text_type(e)}, status=400)
+
+        try:
+            identity_id = data['identity']
+        except KeyError as e:
+            return JsonResponse({'reason': '"identity" must be specified.'},
+                                status=400)
```
rudigiesler: Same comment here as in https://github.com/praekelt/hellomama-registration/pull/139/ about rather user a serializer for this validation.

changes/views.py
```diff
@@ -72,3 +78,54 @@ def post(self, request, *args, **kwargs):
                               action='momconnect_nonloss_optout',
                               data={'reason': 'sms_failure'})
         return Response(status=status.HTTP_201_CREATED)
+
+
+class ReceiveAdminOptout(mixins.CreateModelMixin,
```
rudigiesler: We don't need the `CreateModelMixing` if we're not using the `.create` function.

changes/views.py
```diff
@@ -72,3 +78,54 @@ def post(self, request, *args, **kwargs):
...
+
+        for action in set(actions):
+            Change.objects.create(**{
+                "registrant_id": identity_id,
+                "action": action,
+                "data": {"reason": "other"},
+                "source": source,
+            })
+
+        return JsonResponse({})
```
rudigiesler: We probably want to return a `Response` instead of a `JsonReponse` here, so that Django Rest Framework can handle deciding if the response should be JSON or not, and handle encoding etc..

We should also return with an HTTP 201 result, and probably a list of the new change objects that were created

erikh360:
@rudigiesler I believe I covered everything, please review again.

changes/views.py
```diff
@@ -72,3 +76,62 @@ def post(self, request, *args, **kwargs):
...
+
+        actions = set()
+        for sub in active_subs:
+            messageset = sbm_client.get_messageset(sub["messageset"])
+            if "nurseconnect" in messageset["short_name"]:
+                actions.add("nurse_optout")
+            elif "pmtct" in messageset["short_name"]:
+                actions.add("pmtct_nonloss_optout")
+            else:
+                actions.add("momconnect_nonloss_optout")
```
rudigiesler: I'm not sure that we should consider all other messagesets as momconnect. I think we should just ignore any other messagesets that we don't have an optout type for.

changes/serializers.py
```diff
@@ -11,3 +11,18 @@ class Meta:
...
+
+class MultipleChangeSerializer(serializers.Serializer):
+    changes = ChangeSerializer(many=True)
+
+    def create(self, validated_data):
+        changes_data = validated_data.pop('changes')
+        changes = []
+        for change_data in changes_data:
+            changes.append(Change.objects.create(**change_data))
+        return changes
```
rudigiesler: I don't think that we need the `MultipleChangeSerializer`. I think we can just use `ChangeSerializer(many=True)` directly, and pass it a list instead of the dictionary you're passing this serializer

changes/views.py
```diff
@@ -72,3 +76,62 @@ def post(self, request, *args, **kwargs):
...
+            data["changes"].append(change)
+
+        serializer = MultipleChangeSerializer(data=data)
+
+        if serializer.is_valid():
+            change_objects = serializer.save()
+
+            result = {"results": []}
+            for change in change_objects:
+                result["results"].append(model_to_dict(change))
```
rudigiesler: We can probably just use the `serializer.data` as the results to send back to the user.

PR #112, opened 2017-06-15
Recognise the whatsapp_prebirth message set

smn:


smn:
@erikh360 @rudigiesler @miltontony ready for review

registrations/tasks.py
```diff
@@ -273,8 +274,10 @@ def create_subscriptionrequests(self, registration):
         weeks = 1  # default week number
 
         # . calculate weeks along
-        if (registration.reg_type == "momconnect_prebirth" and
-           registration.source.authority not in ["hw_partial", "patient"]):
+        if (registration.reg_type in [
+                "momconnect_prebirth", "whatsapp_prebirth"] and
+                registration.source.authority not in [
+                    "hw_partial", "patient"]):
```
rudigiesler: I'm not sure that this is the correct place for it. Since we've added the whatsapp choice to the public line, the authority will always be `patient`, so this if statement will never be true for whatsapp registrations.

registrations/models.py
```diff
@@ -49,6 +49,7 @@ class Registration(models.Model):
     REG_TYPE_CHOICES = (
         ('momconnect_prebirth', "MomConnect pregnancy registration"),
         ('momconnect_postbirth', "MomConnect baby registration"),
+        ('whatsapp_prebirth', "WhatsApp MomConnect pregnancy registration"),
```
rudigiesler: I think we might need a migration for something like this? I'm not sure why, since it doesn't change anything database side, but from what I remember django wants a migration for a change like this.

smn: yeah, you're right, adding.

registrations/tasks.py
smn: Yeah so perhaps I should leave it out entirely. I'll do that and see which tests fail, if any.

smn: @rudigiesler so the tests pass but if I leave it out entirely, the weeks always defaults to 1 -- which doen't seem right. Unless that's the case for all public registrations as momconnect_prebirth too?

smn: Reading the ndoh-jsbox code, starting at week 1 does seem to be the case.

rudigiesler (approved):

PR #113, opened 2017-06-20
Allow for multiple nurseconnect messagesets in Changes

KaitCrawford:


rudigiesler (approved):
Not a big thing, I'm happy with the code the way it is, but I think we can simplify some of the loops using `filter` and list comprehension. eg:

```
nc_messagesets = filter(lambda ms: 'nurseconnect' in ms['short_name'], messagesets)
nc_messageset_ids = [ms['id'] for ms in nc_messagesets]
```

rudigiesler (approved):

changes/tasks.py
```diff
@@ -59,35 +59,43 @@ def deactivate_all_except_nurseconnect(self, change):
...
+            active_subs.extend(sbm_client.get_subscriptions(
+                {'identity': change.registrant_id, 'active': True,
+                 'messageset': messageset["id"]}
+            )["results"])
 
         self.l.info("Deactivating active nurseconnect subscriptions")
         for active_sub in active_subs:
             sbm_client.update_subscription(active_sub["id"], {"active": False})
+            # Record the last messageset deactivated to send to jembi
+            change.data["last_messageset"] = active_sub["messageset"]
```
erikh360: How about getting and saving the nurseconnect optout type here instead of in the `get_nurse_optout_type` function? It would save us a trip to SBM.

erikh360 (commented):
Just the one comment.

rudigiesler (approved):

PR #116, opened 2017-06-26
Change language and message endpoint for CI

erikh360:


rudigiesler (commented):
The code itself looks good, I just see two issues.

1. We have a possible race condition. Since we're kicking off two tasks at the same time, they'll probably run at the same time. So it's possible that the language change changes the language of the deactivated subscription before it gets deactivated, and then the new subscription gets created with the old language

2. The current logic for momconnect_change_language changes the languages for ALL momconnect subscriptions. From the API endpoint I would expect it to only change the language of the specified subscription, not all momconnect subscriptions. (the specified subscription could be a non-momconnect one, and then the language doesn't get changed at all)

To avoid these issues, I would suggest the following:
1. Have the new change type change both/either the message set and/or the language of the message set.
2. Do not name the change type `momconnect_`, since it can change any type of subscription, not just momconnect. Something more generic like "change_subscription"

rudigiesler (approved):

PR #118, opened 2017-07-18
Use lang_code for identity language

KaitCrawford:
The language in the Identity store for NdoH is stored under the key `lang_code` not `language`


erikh360 (approved):
Cool!

PR #119, opened 2017-07-27
Additional subscription request for POPI messages

rudigiesler:
What we want to do, is the day after a mother has registered for the service, they should get an SMS detailing how to connect to the new POPI USSD line.

The way this PR proposes to implement it, is to create a new subscription for the mother, to a special message set that contains just the POPI message/s that we want to send, that detail how to access the new POPI line.

### DIFF
--- a/registrations/tasks.py
+++ b/registrations/tasks.py
@@ -265,6 +265,34 @@ class ValidateSubscribe(Task):
             self.l.info("Registration object updated.")
             return False
 
+    def create_popi_subscriptionrequest(self, registration):
+        """
+        Creates a new subscription request for the POPI message set. This
+        message set tells the user how to access the POPI required services.
+        This sould only be sent for Clinic or CHW registrations.
+        """
+        if ('prebirth' not in registration.reg_type or
+           registration.source.authority not in ['hw_partial', 'hw_full']):
+            return "POPI Subscription request not created"
+
+        self.l.info("Fetching messageset")
+        msgset_short_name = utils.get_messageset_short_name(
+            'popi', registration.source.authority, None)
+        msgset_id, msgset_schedule, next_sequence_number =\
+            utils.get_messageset_schedule_sequence(msgset_short_name, None)
+
+        self.l.info("Creating subscription request")
+        from .models import SubscriptionRequest
+        SubscriptionRequest.objects.create(
+            identity=registration.registrant_id,
+            messageset=msgset_id,
+            next_sequence_number=next_sequence_number,
+            lang=registration.data['language'],
+            schedule=msgset_schedule,
+        )
+        self.l.info("POPI Subscription request created")
+        return "POPI Subscription Request created"
+
     # Create SubscriptionRequest
     def create_subscriptionrequests(self, registration):
         """ Create SubscriptionRequest(s) based on the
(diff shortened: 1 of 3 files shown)


registrations/tasks.py
```diff
@@ -265,6 +265,34 @@ def validate(self, registration):
...
+        message set tells the user how to access the POPI required services.
+        This sould only be sent for Clinic or CHW registrations.
+        """
+        if ('prebirth' not in registration.reg_type or
+           registration.source.authority not in ['hw_partial', 'hw_full']):
+            return "POPI Subscription request not created"
+
+        self.l.info("Fetching messageset")
+        msgset_short_name = utils.get_messageset_short_name(
+            'popi', registration.source.authority, None)
```
KaitCrawford: Are we going to create 2 new messagesets then? One for `popi.hw_partial.1` and one for `popi.hw_full.1`?
Perhaps instead we could add a check to `get_messageset_short_name` or `get_messageset_schedule_sequence` that would map the sources to the same messagset?

rudigiesler: Yes, that was the plan.

I thought of modifying the short name function, but the convention for naming the message sets is `<name>.<authority>.<number>`, and I'm not sure what to put in the `<authority>` bit if we make it the same message set, and I'd like to stick to the naming convention.

registrations/tasks.py
```diff
@@ -265,6 +265,34 @@ def validate(self, registration):
             self.l.info("Registration object updated.")
             return False
 
+    def create_popi_subscriptionrequest(self, registration):
+        """
+        Creates a new subscription request for the POPI message set. This
+        message set tells the user how to access the POPI required services.
+        This sould only be sent for Clinic or CHW registrations.
```
erikh360: sould = Should

registrations/tasks.py
```diff
@@ -265,6 +265,34 @@ def validate(self, registration):
...
+        """
+        if ('prebirth' not in registration.reg_type or
```
erikh360: Should it only be for `prebirth`?

rudigiesler: We want to limit it to registrations from the clinic and the CHW lines. Currently, those two USSD lines only do `momconnect_prebirth` and `whatsapp_prebirth` registration types, so combined with limiting the authority type, this should limit it to only those two USSD lines.

registrations/tests.py
```diff
@@ -2473,20 +2475,28 @@ def test_registration_process_clinic_good(self):
         # . check number of calls made:
         #   message set, schedule, jembi registration, id_store mother,
         #   id_store mother_reverse, id_store registrant_rever, id_store patch
-        self.assertEqual(len(responses.calls), 7)
+        self.assertEqual(len(responses.calls), 8)
```
alexmuller: Possibly a nitpick, not sure about this codebase. I think the 7 calls are listed in the comment above, should the 8th one be added somewhere in that list?

rudigiesler: Yup, good call, I'll add it to that list.

alexmuller:
Cool, I think I pretty much understand what's going on here. So the messageset is created in the stage based messaging admin interface and as long as the `MessageSet.short_name` is configured correctly it will get picked up.

rudigiesler:
@alexmuller Yup. We create it in the admin interface with the correct name, and then upload messages using the command line tool: https://github.com/praekelt/seed-services-cli

If it's not configured with the correct name, we'll hopefully get an error picked up by sentry, and we can then fix/add the name in the django admin.

PR #120, opened 2017-08-08
Management command for revalidating invalid registrations

KaitCrawford:
A number of bugs in the system have led to some valid registrations being marked as invalid. These have been fixed so we should try to re-validate the registrations affected. This adds a management command to do that.

### DIFF
--- a/registrations/management/commands/revalidate_registrations.py
+++ b/registrations/management/commands/revalidate_registrations.py
@@ -1,0 +1,102 @@
+from os import environ
+
+from django.core.management.base import BaseCommand, CommandError
+
+from registrations.models import Registration
+from registrations.tasks import (
+    add_personally_identifiable_fields, validate_subscribe)
+
+from seed_services_client import StageBasedMessagingApiClient
+
+from ._utils import validate_and_return_url
+
+
+class Command(BaseCommand):
+    help = ("Validates all invalid Registrations that failed with the given "
+            "error. This should also lead to the creation of a  Subscription "
+            "in the SMB service")
+
+    def add_arguments(self, parser):
+        parser.add_argument(
+            '--blind', action='store_false', default=True,
+            dest='check_subscription',
+            help=('Do not check with the stage based messaging API whether'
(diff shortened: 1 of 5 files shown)


alexmuller (commented):
The code here looks good to me! I probably don't know the codebase well enough to know if it will have any other effects but all seems fine to me.

Just a couple of nitpicky typo comments and a couple of questions.

registrations/management/commands/revalidate_registrations.py
```diff
@@ -0,0 +1,102 @@
...
+            "error. This should also lead to the creation of a  Subscription "
+            "in the SMB service")
+
+    def add_arguments(self, parser):
+        parser.add_argument(
+            '--blind', action='store_false', default=True,
+            dest='check_subscription',
+            help=('Do not check with the stage based messaging API whether'
+                  'or not a subscription for the identity already exists.'
+                  'NOT RECOMMENDED AT ALL'))
```
KaitCrawford: Fair point. I took this from one of the other management commands but I don't think we'd ever want to run this without checking first.

registrations/management/commands/revalidate_registrations.py
```diff
@@ -0,0 +1,102 @@
...
+        for reg in registrations.iterator():
+            self.log("Validating registration %s" % reg.id)
+            if check_subscription and self.count_subscriptions(client, reg):
+                self.log(('Identity %s already has subscription. Skipping.')
+                         % (reg.registrant_id))
+                continue
+            """
+            validate_subscribe() checks all data for the registration is valid
+            and creates the Subscription Request
+            """
```
KaitCrawford: If I'm not mistaken, docstrings mean something to systems that auto-generate your docs. Otherwise I don't think there's any real difference. But I do think it's convention to use docstrings for things that should be documented (like at the top of methods or tests) and comments in the code so I'll change this. Thanks.

alexmuller (approved):
All seems good here! 👍

PR #126, opened 2017-09-28
Add WhatsApp PMTCT registration and changes support

rudigiesler:
We want to add a WhatsApp message sets for the PMTCT message sets, so that those messages can also go over whatsapp.

For this, we need to add support for mothers to register for these new whatsapp message sets, as well as support for changing to loss or postbirth on the whatsapp messagesets.


changes/tests.py
```diff
@@ -1767,6 +1825,75 @@ def test_baby_switch_momconnect_only_sub(self):
...
+        # . mock get subscriptions request
+        sub = mock_get_active_subs_whatsapp(change_data['registrant_id'])
+
+        # . mock get messageset by id
+        utils_tests.mock_get_messageset(99)
+
+        # . mock deactivate active subscriptions
+        mock_deactivate_subscriptions([sub])
+
+        # . mock get momconnect_postrbirh messageset by shortname
```
KaitCrawford: small type `postbirth`

changes/tests.py
```diff
@@ -1823,7 +1950,79 @@ def test_pmtct_loss_switch(self):
...
+            'mha': 1,
+            'swt': 1,
+            'type': 5,
+        })
+
+    @responses.activate
+    def test_pmtct_loss_switch_whatsapp(self):
+        """
+        If the mother was on the whatsapp message set, she should be switched
+        to the whatsapp pmtct message set.
```
KaitCrawford: I think this is supposed to say "to the whatsapp loss message set"?

PR #133, opened 2017-10-30
Fix error handling in Jembi outbound post

KaitCrawford:


registrations/views.py
```diff
@@ -241,14 +241,16 @@ def post(self, request):
...
-            logger.warning(e.response.url)
-            logger.warning(e.response.headers)
-            return Response(
-                'Error when posting to Jembi. Body: %s Payload: %r' % (
-                    e.response.content, post_data),
-                status=status.HTTP_400_BAD_REQUEST)
+            if e.response.status_code == 400:
+                logger.warning("400 Error when posting to Jembi.\n"
+                               "Response: %s\nPayload:%s" %
+                               (e.response.content, post_data))
```
rudigiesler: Getting a python 3 issue here in the tests. `content` is the raw content, and `text` is the decoded content, so we probably want to use `text` instead.

rudigiesler (approved):

rudigiesler (approved):

PR #135, opened 2017-11-15
New Jembi endpoint for NC helpdesk

erikh360:
New link for helpdesk outgoing messages on NC helpdesk:
`/ws/rest/v1/nc/helpdesk`

I use the source object linked to the authenticated user to determine which helpdesk the outgoing comes from.

Payload is the same as Momconnect.

### DIFF
--- a/registrations/views.py
+++ b/registrations/views.py
@@ -204,7 +204,7 @@ class JembiHelpdeskOutgoingView(APIView):
             "repdate": jembi_format_date(
                 validated_data.get('outbound_created_on')),
             "mha": 1,
-            "swt": swt,  # 1 ussd, 2 sms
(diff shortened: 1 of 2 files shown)


registrations/views.py
```diff
@@ -233,8 +233,15 @@ def post(self, request):
 
         post_data = self.build_jembi_helpdesk_json(serializer.validated_data)
         try:
+
+            source = Source.objects.get(user=self.request.user.id)
+
+            endpoint = 'helpdesk'
+            if source.name == 'nc_helpdesk':
```
rudigiesler: Looking at the existing source names in production, I wonder if we can make this more like that format to keep things consistent? 
image

PR #138, opened 2018-01-30
Create change type for switching messaging channel

rudigiesler:
We want users to be able to switch what channel they're currently receiving messages on. To do this, we need to create a new change type. This change type will be able to switch between the current two messaging options, SMS and WhatsApp

### DIFF
--- a/changes/tests.py
+++ b/changes/tests.py
@@ -160,8 +160,9 @@ def mock_get_active_subs_whatsapp(registrant_id):
 
 def mock_get_messagesets(messagesets):
     """
-    Mocks getting the list of messagesets. messagesets is a list of short
-    names
+    Mocks the request for getting the list of messagesets using responses.
+    `messagesets` is a list of short names for the messagesets that should be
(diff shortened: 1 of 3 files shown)


changes/tests.py
```diff
@@ -260,56 +278,54 @@ def mock_get_all_messagesets():
     )
 
 
-def mock_get_active_subs_mc(registrant_id):
-    momconnect_prebirth_sub_id = "subscriptionid-momconnect-prebirth-0"
+def mock_get_subscriptions(querystring=None, results=[]):
+    url = 'http://sbm/api/v1/subscriptions/'
+    if querystring is not None:
+        url = url + querystring
```
nathanbegbie: We can replace this with `url = "http://sbm/api/v1/subscriptions/{}".format(querystring)` which handles the case where querystring is `None`

rudigiesler:
@nathanbegbie Thanks! Made some changes, ready for another review :)

PR #141, opened 2018-03-05
Add Jembi app endpoint

rudigiesler:


registrations/tasks.py
```diff
@@ -503,6 +536,252 @@ def add_personally_identifiable_fields(registration):
...
+    def run(self, registration_id, **kwargs):
+        registration = Registration.objects.get(id=registration_id)
+        msisdn_registrant = registration.data['msisdn_registrant']
+        registrant = self.get_or_update_identity_by_address(msisdn_registrant)
+        device = self.get_or_update_identity_by_address(
+            registration.data['msisdn_device'])
+        registration.registrant_id = registrant['id']
+
+        # Check for existing subscriptions
+        if self.is_identity_subscribed(registrant, r'prebirth\.hw_full'):
```
erikh360, line 711: If this is a pmtct registration wouldn't we need to get past this point?

registrations/test_serializers.py
```diff
@@ -0,0 +1,181 @@
...
+            'mom_dob': '1988-01-01',
+            'mom_lang': 'eng_ZA',
+            'mom_edd': '2016-06-06',
+            'mom_consent': True,
+            'clinic_code': '123456',
+            'mha': 1,
+            'created': '2016-01-01 00:00:00',
+        }
+        serializer = JembiAppRegistrationSerializer(data=data)
+        serializer.is_valid()
```
erikh360: Do we need this line if we call it on the next line again?

registrations/test_tasks.py
```diff
@@ -0,0 +1,864 @@
...
+                "height": 1
+            },
+            match_querystring=True)
+
+        self.assertTrue(task.is_valid_clinic_code('123456'))
+
+    @responses.activate
+    def test_is_valid_clinic_code_false(self):
+        """
+        If there are no results for the clinic code search, then Flase should
```
erikh360: Spelling. False!=Flase

erikh360 (commented):
Looks good! Just a few questions.

registrations/tasks.py
rudigiesler: So, the `pmtct` option on this endpoint is not whether this is a prebirth or pmtct registration, it's whether this is a prebirth or a prebirth + pmtct registration. Either way it's creating a prebirth registration, so we need this check to avoid creating duplicate prebirth subscriptions.

registrations/test_serializers.py
rudigiesler: No, I'll remove it

rudigiesler:
@erikh360 Thanks for the review :) I've pushed some changes, ready for review again

erikh360 (approved):
🍰

PR #142, opened 2018-03-07
Add basic django channels setup

rudigiesler:
In order to test the feasibility of running websockets on our production setup, this PR updates dependancies to be able to run django channels, adds a simple echo endpoint to test the websockets, and updates the docker image to be able to run either gunicorn or daphne, for websocket support.


erikh360 (approved):
Looks good to me.

PR #145, opened 2018-03-09
Add Jembi Registration websocket

rudigiesler:
The registration and registration status views should also be available over websocket, as well as receiving registration events


erikh360 (approved):
Complicated. 🍰

rudigiesler:
@erikh360 Which bits seem complicated? The tests? Testing in django channels seems very complicated :/ I haven't been able to find out a better way yet.

PR #149, opened 2018-03-28
Sending channel switch event to Jembi

erikh360:
This is to send the switch between Whatsapp and SMS to Jembi.

Spec found here: https://docs.google.com/document/d/1080DOh6U2nI0jfw8je6ocVYEiPH5r4PGRHJ6_n6eCFk/edit#heading=h.w5jrvcpwhbs1

I also moved a function used in multiple classes to the Base class.


KaitCrawford (approved):
:+1:

PR #152, opened 2018-04-17
Fix whatsapp prebirth registrations and whatsapp changes

rudigiesler:
Currently, whatsapp prebirth registrations are always for the non-unaccellerated message set, and always at sequence number 1.

Also, baby switch changes are not implemented for whatsapp prebirth subscriptions.

This PR aims to fix both of these bugs


changes/tests.py
```diff
@@ -1942,6 +1952,76 @@ def test_baby_switch_whatsapp(self):
...
+        """
+        If the mother is subscribed to momconnect whatsapp, then when
+        switching to baby messages, they should receive those through
+        whatsapp as well.
+        """
+        # Pretest
+        self.assertEqual(Registration.objects.all().count(), 0)
+        # Setup
+        # make registrations
+        self.make_registration_whatsapp_pmtct_prebirth()
```
KaitCrawford: should this be creating a normal prebirth subscription or a pmtct prebirth subscription?

KaitCrawford (commented):
One question

changes/tests.py
rudigiesler: Yes, I'll fix this. It shouldn't affect the results of the test though, since we're mocking out the fetch subscriptions, but this is confusing to someone reading the test.

rudigiesler:
@KaitCrawford Thanks :) ready for another review

KaitCrawford (approved):
:+1:

PR #155, opened 2018-04-25
Fixes for automatic WhatsApp switch

rudigiesler:
If a token isn't supplied, we previously got an error, because the auth backend returned `False` instead of `None`. This has been fixed.

If we couldn't find an outbound for the ID specified in the event, then previously an error was raised. We should instead just skip the processing of this event


erikh360 (approved):
🍰

PR #156, opened 2018-05-03
Add position tracker

rudigiesler:
For NurseConnect, we want all the nurses to be receiving the same message at the same time, instead of the usual stage based messaging.

For current subscriptions, this is as simple as moving all subscriptions to the same point on the same messageset.

However, we also want new subscriptions to end up at the same place as everyone else.

This position tracker will help with this. Every time there's a new registration, we can look at the position tracker to find out where in the message set we should place the nurse. And then with every message send, we make a request to the API endpoint to increment the tracker position, so that all new registrations get started at the correct point.


registrations/test_views.py
```diff
@@ -111,3 +113,82 @@ def test_get_registration_only_by_user(self):
...
+        url = reverse('positiontracker-list')
+        response = self.normalclient.post(url, data={'label': 'test'})
+        self.assertEqual(response.status_code, 403)
+        self.assertEqual(PositionTracker.objects.count(), 0)
+
+        self.normaluser.user_permissions.add(
+            Permission.objects.get(name='Can add position tracker'))
+
+        response = self.normalclient.post(url, data={'label': 'test'})
+        self.assertEqual(response.status_code, 201)
```
OMosimege: Curious on why is the status_code different here as compared to line 137? Because it looks like the lines leading up to both 137 and line 144 do the same thing. Is it because of the user_permissions.add() in line 140?

OMosimege (approved):
This looks great Rudi! The questions I had soon got answered as I read further down the code :100:

PR #157, opened 2018-05-03
Nurseconnect registration to use position tracker

rudigiesler:
This PR adds a flag, that if set, will ensure that all new nurseconnect registrations are on the RTHB messageset, and that they start in the position according to the position tracker

It also adds a management command to switch over all existing nurseconnect subscriptions to the correct point in the RTHB messageset


erikh360 (approved):
Nice!

PR #159, opened 2018-05-29
Only include primary address when searching for identities by address

rudigiesler:


OMosimege (approved):
Great job Rudi :cake: This looks great

PR #160, opened 2018-07-09
Allow sending optouts to identity store

erikh360:
There is a race condition in the functionality that creates the optout changes from the popi jsbox app here:
https://github.com/praekeltfoundation/ndoh-jsbox/blob/develop/src/ussd_popi_user_data.js#L679-L703 

We create these changes from the jsbox apps and simultaneously send the optout to the identity store.

The change tries to send the details to Jembi but by then the identity optout has already happened and there is no msisdn on the identity. The result is that we send a empty msisdn to Jembi.

This will only optout and remove the msisdn after the request was sent to Jembi. The jsbox app will also be updated.

### DIFF
--- a/changes/tasks.py
+++ b/changes/tasks.py
@@ -1015,6 +1015,9 @@ class BasePushOptoutToJembi(object):
                 verify=False
             )
             result.raise_for_status()
+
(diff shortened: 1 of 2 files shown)


DevChima (commented):
+1

PR #161, opened 2018-08-08
Add welcome message for jembi android app registrations

rudigiesler:
For the USSD registrations, the USSD app makes the API call to the message sender to send the welcome message. Jembi's android app, however, cannot do this, so we must add it as part of the registration task.

This PR adds translation support, so that the message is in the correct language, adds the message sending as part of the registration task, as well as some travis config fixes, and seaworthy tests for the docker image

### DIFF
--- a/.dockerignore
+++ b/.dockerignore
@@ -1,1 +1,103 @@
-.git
+# Docker-specific things:
+# Docker files
+Dockerfile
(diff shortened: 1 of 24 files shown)


OMosimege (approved):
Great Job Rudi! 🍰

PR #162, opened 2018-08-20
Create subscriptions for service info messages for new registrations

rudigiesler:
This PR creates subscriptions to the new service info messages for all new whatsapp registrations, placing them in the correct position of the service info messageset according to their EDD.


OMosimege (approved):
Looks great! 🍰

PR #164, opened 2018-09-04
Add metadata field to welcome outbound message send

rudigiesler:
The request to create the welcome message is missing the metadata field, which is resulting in a failed request.


erikh360 (approved):
Cool! when travis is happy. 🍰

PR #165, opened 2018-10-01
Handle all engage failed types

erikh360:
This only updates the endpoint to work with the new engage format and auth.

Still need to handle each event properly and send notification.

### DIFF
--- a/changes/serializers.py
+++ b/changes/serializers.py
@@ -47,14 +47,14 @@ class AdminChangeSerializer(serializers.Serializer):
 
 class ReceiveWhatsAppEventSerializer(serializers.Serializer):
 
-    class HookSerializer(serializers.Serializer):
-        event = serializers.ChoiceField(['message.direct_outbound.status'])
-    hook = HookSerializer()
+    class StatusSerializer(serializers.Serializer):
 
-    class EventSerializer(serializers.Serializer):
-        status = serializers.ChoiceField(['unsent'])
+        class ErrorSerializer(serializers.Serializer):
+            code = serializers.CharField(required=False)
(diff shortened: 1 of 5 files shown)


changes/serializers.py
```diff
@@ -47,14 +47,14 @@ class AdminChangeSerializer(serializers.Serializer):
...
+        class ErrorSerializer(serializers.Serializer):
+            code = serializers.CharField(required=False)
+            title = serializers.CharField(required=True)
 
-        class MessageMetadata(serializers.Serializer):
-            junebug_message_id = serializers.UUIDField()
-        message_metadata = MessageMetadata()
-    data = EventSerializer()
+        id = serializers.CharField(required=True)
+        status = serializers.ChoiceField(['failed'])
```
rudigiesler: I think we should have a test to confirm this, because I'm not 100% sure how this will work, but what if in the list of statuses that we're given, some of them are `failed`, and some of them are `succeeded`? Will the entire request fail with a 400, or will some of the statuses go through, and some of them fail? I'm not sure how this works with ListFields.

Also, if it does go through because some of them match, we probably need to filter it in the view to only action on the failed ones (unless the serializer does this for us?).

Either way, I think we should have a test to confirm that this works the way we expect it to.

rudigiesler (commented):
Just one comment around that I'm unsure how ListFields work, the rest looks good :)

changes/serializers.py
erikh360: Good catch, I'll write a test to see what happens and handle accordingly.

erikh360:
@rudigiesler ready for another look.

rudigiesler (approved):

PR #166, opened 2018-10-02
Handle HSM errors appropriately

erikh360:
This change will only do the switch to SMS on the correct errors, time based failures will be dealt with in another endpoint.

I still need to do the translations.

### DIFF
--- a/changes/serializers.py
+++ b/changes/serializers.py
@@ -50,7 +50,7 @@ class ReceiveWhatsAppEventSerializer(serializers.Serializer):
     class StatusSerializer(serializers.Serializer):
 
         class ErrorSerializer(serializers.Serializer):
-            code = serializers.CharField(required=False)
+            code = serializers.IntegerField(required=False)
             title = serializers.CharField(required=True)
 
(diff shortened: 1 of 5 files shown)


changes/serializers.py
```diff
@@ -50,7 +50,7 @@ class ReceiveWhatsAppEventSerializer(serializers.Serializer):
     class StatusSerializer(serializers.Serializer):
 
         class ErrorSerializer(serializers.Serializer):
-            code = serializers.CharField(required=False)
+            code = serializers.IntegerField(required=False)
```
rudigiesler, line 53: We'll probably need a migration for this?

rudigiesler (commented):
Just one small comment, the rest of it looks good

changes/serializers.py
erikh360: Not for a serializer, do we?

rudigiesler: Oh sorry, thought this was a model for some strange reason, 👍

rudigiesler (approved):

PR #168, opened 2018-10-04
Add system event handler

erikh360:
This PR adds a system event handler coming from engage.

For now we are only handling `undelivered` which means a message was not delivered in a week and no messages sent after it has been delivered.

We use this to notify the user that they can switch to SMS.

Translation support is there but we don't have the translations yet.

### DIFF
--- a/changes/serializers.py
+++ b/changes/serializers.py
@@ -71,3 +71,13 @@ class ReceiveWhatsAppEventSerializer(serializers.Serializer):
 
         return super(ReceiveWhatsAppEventSerializer, self).to_internal_value(
             data)
+
+
+class ReceiveWhatsAppSystemEventSerializer(serializers.Serializer):
+
+    class EventSerializer(serializers.Serializer):
+        type = serializers.CharField(required=True)
+        message_id = serializers.CharField(required=True)
+        recipient_id = serializers.CharField(required=True)
+
+    events = serializers.ListField(child=EventSerializer(), min_length=1)
--- a/changes/test_tasks.py
+++ b/changes/test_tasks.py
@@ -205,6 +205,9 @@ class ProcessWhatsAppSystemEventTaskTests(WhatsAppBaseTestCase):
         """
         self.create_outbound_lookup(0)
 
-        process_whatsapp_system_event('messageid', "something_else")
+        process_whatsapp_system_event('messageid', "undelivered")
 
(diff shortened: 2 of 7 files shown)


changes/serializers.py
```diff
@@ -71,3 +71,13 @@ def to_internal_value(self, data):
 
         return super(ReceiveWhatsAppEventSerializer, self).to_internal_value(
             data)
+
+
+class ReceiveWhatsAppSystemEventSerializer(serializers.Serializer):
+
+    class EventSerializer(serializers.Serializer):
+        type = serializers.CharField(required=True)
```
erikh360, line 79: I made this a `CharField` instead of a `ChoiceField` because new types can be added to Engage at any time.

changes/test_tasks.py
```diff
@@ -168,3 +159,52 @@ def test_non_hsm_failure(self):
...
+    @mock.patch('changes.tasks.utils.ms_client.create_outbound')
+    @responses.activate
+    def test_no_message_found(self, mock_create_outbound):
+        """
+        The task should not create a outbound if the original message was not
+        found
+        """
+        self.create_outbound_lookup(0)
+
+        process_whatsapp_system_event('messageid', "something_else")
```
rudigiesler: Shouldn't this be an `undelivered` type, so that we're actually moving on to the message processing, rather than exiting at the event type handling?

We should probably add a check to this test (maybe ensure that the outbound request was actually made) to ensure that we're actually getting that far in the processing.

changes/views.py
```diff
@@ -237,3 +240,20 @@ def post(self, request, *args, **kwargs):
...
+
+class ReceiveWhatsAppSystemEvent(ReceiveWhatsAppBase):
+    serializer_class = ReceiveWhatsAppSystemEventSerializer
+
+    def post(self, request, *args, **kwargs):
+        self.validate_signature(request)
+        serializer = self.get_serializer(data=request.data)
+        if not serializer.is_valid():
+            # If this isn't an event that we care about
+            return Response(status=status.HTTP_204_NO_CONTENT)
```
rudigiesler: Shouldn't this rather be a 400 error? Since we're accepting all event types, if the serializer fails, it means that the request is in an invalid format, not that this isn't an event that we care about.

rudigiesler (approved):

PR #169, opened 2018-10-05
Add webhook receiver for message sender events

rudigiesler:
This PR adds a view to receive webhook events from the Seed Message Sender that were added in https://github.com/praekelt/seed-message-sender/pull/96

It currently supports the whatsapp contact check failure event.


OMosimege (approved):
👍

PR #170, opened 2018-10-15
Upgrade packages

rudigiesler:
This PR upgrades Django, and all our other dependancies, to the latest version.

We are not upgrade to Python 3.7 yet, as celery does not yet support 3.7


OMosimege (approved):
Great work! 🍰

PR #172, opened 2018-10-17
Add Engage Context endpoint

rudigiesler:
This PR adds the view for the engage webhook to call to get the context to display in the UI.

It currently returns registration and identity information. Subscription information will be added in another PR.


OMosimege (approved):
Looks great! 👍

registrations/views.py
```diff
@@ -625,3 +632,140 @@ def increment_position(self, request, pk=None):
...
+                    "icon": "info-circle",
+                    "title": "Mother's Details",
+                    "type": "table",
+                    "timestamp": timezone.now().isoformat(),
+                    "uuid": "7eb448be-3180-417e-823c-0c6d5da24e00",
+                    "payload": info,
+                }
+            )
+
+        return Response({"version": "1.0.0-alpha", "context": context})
```
erikh360, line 771: Not sure about this version? Shouldn't it be more dynamic.

erikh360 (commented):
Looks good, just one question.

registrations/views.py
rudigiesler: This is us sending it to engage, according to the API docs.

I'm not entirely sure what it is for, but my guess is that if engage ever changes this API, then it will identify and understand the old format by this string, and then when we upgrade to the new format we change this string to show that we're now sending in the new format, basically kind of like a "message format". So we want it to be static, if I'm understanding correctly.

erikh360: Oh cool.

PR #174, opened 2018-10-22
Feature/momza 1062 move to engage api

OMosimege:
change over from WASSUP to ENGAGE API

### DIFF
--- a/registrations/tasks.py
+++ b/registrations/tasks.py
@@ -708,13 +708,14 @@ class ValidateSubscribeJembiAppRegistration(HTTPRetryMixin, ValidateSubscribe):
         Returns whether or not the number is recognised on wassup
         """
         r = requests.post(
-            settings.WASSUP_URL,
-            json={"number": settings.WASSUP_NUMBER, "msisdns": [address], "wait": True},
-            headers={"Authorization": "Token {}".format(settings.WASSUP_TOKEN)},
+            settings.ENGAGE_URL,
+            json={"blocking": "wait", "contacts": [address]},
+            headers={"Authorization": "Bearer {}".format(settings.ENGAGE_TOKEN)},
         )
         r.raise_for_status()
         data = r.json()
-        existing = filter(lambda d: d.get("status", False) == "valid", data)
+        existing = filter(lambda d: d.get('status', False) == "valid",
+                          data[0]['contacts'])
         return any(existing)
 
     def create_pmtct_registration(self, registration, operator):
--- a/registrations/tasks.py
+++ b/registrations/tasks.py
@@ -710,11 +710,7 @@ class ValidateSubscribeJembiAppRegistration(HTTPRetryMixin, ValidateSubscribe):
         r = requests.post(
             urljoin(settings.ENGAGE_TOKEN, "http://engage/v1/contacts"),
             json={"blocking": "wait", "contacts": [address]},
-            headers={
-                "Authorization": "Bearer {}".format(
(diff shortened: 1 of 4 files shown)


registrations/tasks.py
```diff
@@ -708,13 +708,14 @@ def is_registered_on_whatsapp(self, address):
         Returns whether or not the number is recognised on wassup
         """
         r = requests.post(
-            settings.WASSUP_URL,
-            json={"number": settings.WASSUP_NUMBER, "msisdns": [address], "wait": True},
-            headers={"Authorization": "Token {}".format(settings.WASSUP_TOKEN)},
+            settings.ENGAGE_URL,
```
rudigiesler: I think we should move away from specifying the exact URL endpoint for the contact check (eg `whatsapp.praekelt.org/v1/contacts`), and move towards specifying the base url (eg `whatsapp.praekelt.org`), and then using `urljoin` (https://docs.python.org/3.5/library/urllib.parse.html#urllib.parse.urljoin) to get the specific path.

Right now, because this is the only endpoint, it's not necessary, but I think we should plan for the future for when we want to access different endpoints on the engage API.

registrations/test_tasks.py
```diff
@@ -297,21 +297,24 @@ def test_is_registered_on_whatsapp_true(self):
...
             self.assertEqual(
-                data,
-                {"number": "+27820000000", "msisdns": ["+27821111111"], "wait": True},
+                data, {"blocking": "wait", "contacts": ["+27821111112"]},
             )
-            return (200, {}, json.dumps([{"status": "valid", "wa_id": "27821111111"}]))
+            return (200, {}, json.dumps([{"contacts": [{"input": "+27821111112",
+                                                        "status": "valid",
+                                                        "wa_id": "27821111112"}]
+                                          }]))
```
rudigiesler: The Engage API returns an object, not an array, eg. `{"contacts": [...]}` not `[{"contacts": [...]}]`

registrations/tasks.py
OMosimege: so I should not use `settings.ENGAGE_URL`?

rudigiesler: I think we should be using `settings.ENGAGE_URL`, but we should assume that `ENGAGE_URL` is something like `whatsapp.praekelt.org`, not the full path of `whatsapp.praekelt.org/v1/contacts`, and that we need to add `v1/contacts` to `ENGAGE_URL` to be able to get the full URL that we need to call for the contact check.

registrations/tasks.py
```diff
@@ -708,13 +708,13 @@ def is_registered_on_whatsapp(self, address):
         Returns whether or not the number is recognised on wassup
         """
         r = requests.post(
-            settings.WASSUP_URL,
-            json={"number": settings.WASSUP_NUMBER, "msisdns": [address], "wait": True},
-            headers={"Authorization": "Token {}".format(settings.WASSUP_TOKEN)},
+            urljoin(settings.ENGAGE_TOKEN, "http://engage/v1/contacts"),
```
rudigiesler:
```suggestion
            urljoin(settings.ENGAGE_URL, "v1/contacts"),
```

rudigiesler (changes requested):

registrations/tasks.py
OMosimege: thanks Rudi, missed that oversight

rudigiesler (approved):

PR #176, opened 2018-10-30
Switch to sms fix

erikh360:
SEED should not try and create a SMS service info subscription when user changes channel from WA to SMS.

### DIFF
--- a/changes/tasks.py
+++ b/changes/tasks.py
@@ -545,15 +545,19 @@ class ValidateImplement(Task):
             elif change.data["channel"] == "sms" and "whatsapp" in short_name:
                 # Change any WhatsApp subscriptions to SMS
                 sbm_client.update_subscription(sub["id"], {"active": False})
-                if "whatsapp_service_info" not in short_name:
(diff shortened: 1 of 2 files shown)


changes/tasks.py
```diff
@@ -545,14 +545,15 @@ def switch_channel(self, change):
...
-                )
+                if "whatsapp_service_info" not in short_name:
+                    messageset = messagesets_rev[re.sub("^whatsapp_", "", short_name)]
+                    SubscriptionRequest.objects.create(
+                        identity=sub["identity"],
+                        messageset=messageset,
+                        next_sequence_number=sub["next_sequence_number"],
+                        lang=sub["lang"],
+                        schedule=sub["schedule"],
+                    )
```
rudigiesler: Maybe we can rather do a:
```
# There's no service info for SMS
if "service_info" in short_name:
    continue
```
To firstly better explain it, and secondly avoid nesting things too deeply.

rudigiesler (approved):

PR #177, opened 2018-10-30
Switch to whatsapp fix

erikh360:
SEED should create a service info subscription when a user switches from SMS to WhatsApp

### DIFF
--- a/changes/tasks.py
+++ b/changes/tasks.py
@@ -551,9 +551,8 @@ class ValidateImplement(Task):
                 msgset_short_name = utils.get_messageset_short_name(
                     "whatsapp_service_info", reg.source.authority, weeks
                 )
-                msgset_id, msgset_schedule, next_sequence_number = utils.get_messageset_schedule_sequence(
-                    msgset_short_name, weeks
-                )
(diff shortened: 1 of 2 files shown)


changes/tasks.py
```diff
@@ -542,6 +542,26 @@ def switch_channel(self, change):
...
+                r = utils.get_messageset_schedule_sequence(msgset_short_name, weeks)
+                msgset_id, msgset_schedule, next_sequence_number = r
+
+                SubscriptionRequest.objects.create(
+                    identity=reg.registrant_id,
+                    messageset=msgset_id,
+                    next_sequence_number=next_sequence_number,
+                    lang=sub["lang"],
+                    schedule=msgset_schedule,
+                )
```
rudigiesler, line 568: This seems to run for every active SMS subscription, which means if the mother has both a `momconnect_prebirth` and `pmtct_prebirth` subscriptions, then she'll get two subscriptions to the service info message set.

We should probably only create the service info subscription for each momconnect pre/post birth message set

erikh360: Good catch, thanks!

rudigiesler (approved):

PR #181, opened 2018-11-02
No sms messageset notification

erikh360:
Mom who is on 1-2 Message Set (WA) and tries to switch herself to SMS will get the following autoresponse: _“We notice that you have been receiving MomConnect msgs on WhatsApp for children between 1 - 2. Messages for children between 1 - 2 are only available on WhatsApp - switching to SMS means you will not receive any messages. You can stop your MomConnect messages completely by replying “STOP”._ *NB: The system MUST NOT switch the mom to SMS, she can either stay on WA or initiate an opt out herself.*

### DIFF
--- a/changes/tasks.py
+++ b/changes/tasks.py
@@ -577,9 +577,7 @@ class ValidateImplement(Task):
                 if "service_info" in short_name:
                     continue
 
-                messageset = messagesets_rev.get(re.sub("^whatsapp_", "", short_name))
-
-                if not messageset:
+                if short_name == "whatsapp_momconnect_postbirth.hw_full.3":
                     text = (
                         "We notice that you have been receiving MomConnect msgs on "
                         "WhatsApp for children between 1 - 2. Messages for children "
@@ -602,6 +600,7 @@ class ValidateImplement(Task):
 
                 # Change any WhatsApp subscriptions to SMS
                 sbm_client.update_subscription(sub["id"], {"active": False})
+                messageset = messagesets_rev[re.sub("^whatsapp_", "", short_name)]
 
                 SubscriptionRequest.objects.create(
                     identity=sub["identity"],
--- a/changes/tasks.py
+++ b/changes/tasks.py
@@ -590,7 +590,7 @@ class ValidateImplement(Task):
                         {
                             "to_identity": change.registrant_id,
                             "content": text,
-                            "channel": "JUNE_TEXT",
+                            "channel": "WHATSAPP",
                             "metadata": {},
                         }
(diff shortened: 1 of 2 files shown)


changes/tasks.py
```diff
@@ -572,14 +572,37 @@ def switch_channel(self, change):
...
-                sbm_client.update_subscription(sub["id"], {"active": False})
 
                 # There's no service info for SMS
                 if "service_info" in short_name:
                     continue
 
-                messageset = messagesets_rev[re.sub("^whatsapp_", "", short_name)]
+                messageset = messagesets_rev.get(re.sub("^whatsapp_", "", short_name))
+
+                if not messageset:
```
rudigiesler: I'm worried that we're including too much here. It would rather be better that if we forget about any messageset that's whatsapp only, that it results in an exception, rather than sending the message that is very specific for 1-2 messaging, and us not noticing that anything is wrong.

So I think we should rather check if the message is the specific messageset that we're concerned about here, rather than checking for any missing messageset.

changes/tasks.py
```diff
@@ -572,14 +572,37 @@ def switch_channel(self, change):
...
+                        "between 1 - 2 are only available on WhatsApp - switching to "
+                        "SMS means you will not receive any messages. You can stop "
+                        "your MomConnect messages completely by replying ‘STOP‘"
+                    )
+
+                    utils.ms_client.create_outbound(
+                        {
+                            "to_identity": change.registrant_id,
+                            "content": text,
+                            "channel": "JUNE_TEXT",
```
rudigiesler: This should be sent over WhatsApp

changes/tasks.py
```diff
@@ -572,14 +572,37 @@ def switch_channel(self, change):
...
+                            "metadata": {},
+                        }
+                    )
+                    change.data["error"] = "No whatsapp available"
```
rudigiesler: I don't think that this is descriptive enough. Maybe something like "WhatsApp-only messagesets cannot be switched to SMS"

changes/tasks.py
```diff
@@ -572,14 +572,38 @@ def switch_channel(self, change):
                 )
 
             elif change.data["channel"] == "sms" and "whatsapp" in short_name:
-                # Change any WhatsApp subscriptions to SMS
-                sbm_client.update_subscription(sub["id"], {"active": False})
```
rudigiesler: This means that we'll no longer be deactivating the service info messageset, which we still want to do.

rudigiesler (approved):

PR #182, opened 2018-11-02
Send engage helpdesk replies to DHIS2

rudigiesler:
This PR modifies the engage webhook view to also accept outbound + filter outbound messages, and adds tasks to process this information and send the relevant information to OpenHIM to be placed in DHIS2.

This relies on the `author` field on the engage API, which has not yet been implemented.


OMosimege (approved):
Great work! 🍰

OMosimege (approved):
everything looks good 👍

PR #183, opened 2018-11-08
Management command for TeenMomConnect post birth subscriptions

rudigiesler:
This PR adds a management command that takes an xlsx document, which is an export from RapidPro, and creates subscriptions to the post birth messageset for those mothers.

It has protection against invalid values for any of the fields that we require, and it won't create a subscription if the mother already has an active subscription, making it idempotent

### DIFF
--- a/registrations/management/commands/tests/test_teenmomconnect_postbirth_subscriptions.py
+++ b/registrations/management/commands/tests/test_teenmomconnect_postbirth_subscriptions.py
@@ -25,8 +25,8 @@ def randomword(length):
 
 
 class TeenMomConnectPostbirthSubscriptionsCommandTests(TestCase):
-    @classmethod
-    def generate_workbook(cls, contacts):
+    @staticmethod
+    def generate_workbook(contacts):
         """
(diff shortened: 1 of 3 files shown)


registrations/management/commands/tests/test_teenmomconnect_postbirth_subscriptions.py
```diff
@@ -0,0 +1,377 @@
...
+        self.assertIn(
+            "Invalid language code arg. Skipping...", command.stdout.getvalue()
+        )
+
+    @responses.activate
+    def test_create_or_update_identity_exists(self):
+        """
+        If the identity exists, and the language code is correct, it should be returned
+        """
+        identity_details = {
```
erikh360: I see this happens in a few places, can we move it into functions for more DRY approach?

erikh360 (commented):
Cool!

Just the one comment for the identity mock.

rudigiesler:
@erikh360 Done :) Ready for another review

erikh360 (approved):
Awesome!

PR #184, opened 2018-11-14
Move location of from address in whatsapp webhook request

rudigiesler:
The location of the from address in the webhook has been changed, so we need to update our side to reflect that change.


OMosimege (approved):
looks great! 🍰

PR #187, opened 2018-11-16
Ensure that operator field is a string that contains an integer

rudigiesler:
So even though Jembi want us to send an integer, their validation is for a string. So we need to send a string whose contents are an integer for them to accept it


erikh360 (approved):
hehe

PR #189, opened 2018-11-27
Feature/remove channel settings

OMosimege:
This PR has removed the ndoh_hub/asgi.py, ndoh_hub/routing.py, registrations/consumers.py files, and 
removed `channels` from setup.py


ndoh_hub/settings.py
```diff
@@ -46,7 +46,6 @@
     "django.contrib.messages",
     "django.contrib.staticfiles",
     # 3rd party
-    "channels",
     "raven.contrib.django.raven_compat",
     "rest_framework",
```
rudigiesler, line 50: There's also a `CHANNEL_LAYERS` value in `settings.py` that we can remove

registrations/test_tasks.py
```diff
@@ -168,64 +168,6 @@ def test_opt_in(self):
             },
         )
 
-    @responses.activate
-    @mock.patch("registrations.tasks.group_send")
-    def test_send_webhook(self, websocket):
-        """
-        Sends a webhook to the specified URL with the registration status
-        Also send the status over websocket
-        """
```
rudigiesler: I don't think that we want to remove these tests completely, since it is `Sends a webhook ... Also send the status over websocket`, we just want to remove the websocket things, we still want to test that the webhooks work.

setup.py
```diff
@@ -28,9 +28,6 @@
         # The latest twisted was released without a wheel distribution, so it
         # required gcc to install, which broke our docker build. So we pin it
         # here until a release with a wheel is created
```
rudigiesler: We can also remove this comment now that we're removing the Twisted dependancy

ndoh_hub/settings.py
OMosimege: Thanks will remove that as well

registrations/test_tasks.py
OMosimege: The test mock uses `group_send` which I have deleted, which then caused the test to fail. I'll assume that the next step would be to to replace the mock.patch, but I'm not sure with what

rudigiesler: Since `group_send` isn't being used anymore, we can just remove that mock from the test, and remove the assertions on that mock, and it then should pass, since nothing is calling that mock anymore

registrations/test_tasks.py
```diff
@@ -169,7 +169,9 @@ def test_opt_in(self):
         )
 
     @responses.activate
-    @mock.patch("registrations.tasks.group_send")
+    @mock.patch(
+        "registrations.tasks.validate_subscribe_jembi_app_registration." "send_webhook"
```
rudigiesler: We can make this one string

registrations/test_tasks.py
```diff
@@ -193,39 +195,6 @@ def test_send_webhook(self, websocket):
...
-        user = User.objects.create_user("test", "test@example.org", "test")
-        source = Source.objects.create(
-            name="testsource", user=user, authority="hw_full"
-        )
-        reg = Registration.objects.create(
-            reg_type="jembi_momconnect", source=source, data={}, created_by=user
-        )
-
-        task.send_webhook(reg)
-        self.assertEqual(len(responses.calls), 0)
```
rudigiesler: I think we still need this test. This is testing that if no URL is specified, then it doesn't break things, it should just not make the request for the webhook

registrations/test_tasks.py
rudigiesler: I also don't think that this is necessary? Since we're calling the `send_webhook` function directly, we don't want to mock it out

registrations/test_tasks.py
```diff
@@ -193,39 +195,6 @@ def test_send_webhook(self, websocket):
 
         task.send_webhook(reg)
 
-        self.assertEqual(json.loads(responses.calls[-1].request.body), reg.status)
-        self.assertEqual(
-            responses.calls[-1].request.headers["Authorization"], "Bearer test-token"
-        )
```
rudigiesler: We still want these assertions, we just don't want the websocket assertions

rudigiesler (approved):

PR #190, opened 2018-11-29
Feature/momza 1044 add wa message expiry event

OMosimege:
We want to ensure that moms/nurses always get their messages.

However, even if a user has a device that supports HSM messages, they may still delete their WhatsApp, or not have airtime/data.

WhatsApp has a TTL on templated messages that we can use for this.

The solution that was decided on this:

If a WhatsApp isn't delivered for x time period, we send them an SMS telling them that
We only send an SMS once every y time period
eg. If x is 7 days and y is a month, and there's a message that hasn't been delivered in 7 days, we tell them that. Any other failed messages within a month from that time, we do nothing with. If after a month, there'

### DIFF
--- a/changes/tasks.py
+++ b/changes/tasks.py
@@ -1,7 +1,7 @@
 import json
 import re
 import time
-from datetime import datetime, timedelta
+from datetime import datetime
 from itertools import chain as ichain
 from itertools import dropwhile, takewhile
 from uuid import UUID
--- a/changes/tasks.py
+++ b/changes/tasks.py
@@ -1,4 +1,5 @@
-import datetime
+import time
+from datetime import datetime, timedelta
 import json
 import re
 from itertools import chain as ichain
@@ -1047,7 +1048,7 @@ def restore_personally_identifiable_fields(change):
 
 class BasePushOptoutToJembi(object):
     def get_today(self):
-        return datetime.datetime.today()
+        return datetime.today()
 
     def get_timestamp(self, change):
         return change.created_at.strftime("%Y%m%d%H%M%S")
@@ -1468,6 +1469,64 @@ class ProcessWhatsAppSystemEvent(Task):
 process_whatsapp_system_event = ProcessWhatsAppSystemEvent()
 
 
+class ProcessWhatsAppTimeoutSystemEvent(Task):
+    """
+    Notifies a user that a message we send them on Whatsapp has not been delivered
+    within 7 days, she will be sent an outbound
+    message informing them and give the option to switch to SMS.
+    """
+
+    name = "ndoh_hub.changes.tasks.process_whatsapp_timeout_system_event"
+
+    def handle_undelivered(self, identity_uuid):
+        identity = is_client.get_identity(identity_uuid)
+        # Transform to django language code
+        language = identity["details"]["lang_code"].lower().replace("_", "-")
+        with translation.override(language):
+            text = translation.ugettext(
+                "We see that your MomConnect WhatsApp messages are not being "
+                "delivered. If you would like to receive your messages over "
+                "SMS, reply ‘SMS’."
+            )
+
+        utils.ms_client.create_outbound(
+            {
+                "to_identity": identity_uuid,
+                "content": text,
+                "channel": "JUNE_TEXT",
+                "metadata": {},
+            }
+        )
+
+    def run(self, vumi_message_id: str, timestamp: int, event_type: str,
+            **kwargs) -> None:
+        try:
+            identity_uuid: str = next(
+                utils.ms_client.get_outbounds({"vumi_message_id": vumi_message_id})[
+                    "results"
+                ]
+            )["to_identity"]
+        except StopIteration:
+            """
+            Outbound with message id doesn't exist, so don't continue
+            """
+            return
+
+        if event_type == "undelivered":
+            d1 = datetime.fromtimestamp(timestamp)
+            d2 = datetime.today()
+            week1 = (d1 - timedelta(days=d1.weekday()))
+            week2 = (d2 - timedelta(days=d2.weekday()))
+            # Returns 0 if both dates fall withing one week, 1 if on two weeks etc.
+            weeks = int(round((week2 - week1).days / 7))
+            months = int(round((week2 - week1).days / 30))
+            if (weeks == 1 and months == 0):
+                self.handle_undelivered(identity_uuid)
+
+
+process_whatsapp_timeout_system_event = ProcessWhatsAppTimeoutSystemEvent()
+
+
 class ProcessWhatsAppContactCheckFail(Task):
     """
     Switches the user back to SMS if they don't exist on the WhatsApp network
@@ -1633,8 +1692,8 @@ def get_identity_from_msisdn(context, field):
 
 @app.task
 def send_helpdesk_response_to_dhis2(context):
-    encdate = datetime.datetime.utcfromtimestamp(int(context["inbound_timestamp"]))
-    repdate = datetime.datetime.utcfromtimestamp(int(context["reply_timestamp"]))
+    encdate = datetime.utcfromtimestamp(int(context["inbound_timestamp"]))
+    repdate = datetime.utcfromtimestamp(int(context["reply_timestamp"]))
 
     msisdn = phonenumbers.parse(context["inbound_address"], "ZA")
     msisdn = phonenumbers.format_number(msisdn, phonenumbers.PhoneNumberFormat.E164)
--- a/changes/tasks.py
+++ b/changes/tasks.py
@@ -1498,8 +1498,9 @@ class ProcessWhatsAppTimeoutSystemEvent(Task):
             }
         )
 
-    def run(self, vumi_message_id: str, timestamp: int, event_type: str,
-            **kwargs) -> None:
+    def run(
+        self, vumi_message_id: str, timestamp: int, event_type: str, **kwargs
+    ) -> None:
         try:
             identity_uuid: str = next(
                 utils.ms_client.get_outbounds({"vumi_message_id": vumi_message_id})[
@@ -1515,12 +1516,12 @@ class ProcessWhatsAppTimeoutSystemEvent(Task):
         if event_type == "undelivered":
             d1 = datetime.fromtimestamp(timestamp)
             d2 = datetime.today()
-            week1 = (d1 - timedelta(days=d1.weekday()))
-            week2 = (d2 - timedelta(days=d2.weekday()))
+            week1 = d1 - timedelta(days=d1.weekday())
+            week2 = d2 - timedelta(days=d2.weekday())
             # Returns 0 if both dates fall withing one week, 1 if on two weeks etc.
             weeks = int(round((week2 - week1).days / 7))
             months = int(round((week2 - week1).days / 30))
-            if (weeks == 1 and months == 0):
+            if weeks == 1 and months == 0:
                 self.handle_undelivered(identity_uuid)
 
 
--- a/changes/tasks.py
+++ b/changes/tasks.py
@@ -1,6 +1,5 @@
 import json
 import re
-import time
 from datetime import datetime, timedelta
 from itertools import chain as ichain
 from itertools import dropwhile, takewhile
@@ -1482,6 +1481,7 @@ class ProcessWhatsAppTimeoutSystemEvent(Task):
         identity = is_client.get_identity(identity_uuid)
         # Transform to django language code
         language = identity["details"]["lang_code"].lower().replace("_", "-")
+        identity["details"]["timeout_timestamp"] = datetime.today()
         with translation.override(language):
             text = translation.ugettext(
                 "We see that your MomConnect WhatsApp messages are not being "
@@ -1499,7 +1499,12 @@ class ProcessWhatsAppTimeoutSystemEvent(Task):
         )
 
     def run(
-        self, vumi_message_id: str, timestamp: int, event_type: str, **kwargs
+        self,
+        vumi_message_id: str,
+        timestamp: int,
+        event_type: str,
+        timeout_timestamp: int,
+        **kwargs
     ) -> None:
         try:
             identity_uuid: str = next(
@@ -1514,15 +1519,18 @@ class ProcessWhatsAppTimeoutSystemEvent(Task):
             return
 
         if event_type == "undelivered":
-            d1 = datetime.fromtimestamp(timestamp)
-            d2 = datetime.today()
-            week1 = d1 - timedelta(days=d1.weekday())
-            week2 = d2 - timedelta(days=d2.weekday())
-            # Returns 0 if both dates fall withing one week, 1 if on two weeks etc.
-            weeks = int(round((week2 - week1).days / 7))
-            months = int(round((week2 - week1).days / 30))
-            if weeks == 1 and months == 0:
+            if timeout_timestamp is None:
                 self.handle_undelivered(identity_uuid)
+            else:
+                d1 = datetime.strptime(timeout_timestamp, "%d/%m/%Y")
+                d2 = datetime.today()
+                week1 = d1 - timedelta(days=d1.weekday())
+                week2 = d2 - timedelta(days=d2.weekday())
+                # Returns 0 if both dates fall withing one week, 1 if on two weeks etc.
+                weeks = int(round((week2 - week1).days / 7))
+                months = int(round((week2 - week1).days / 30))
+                if weeks == 1 and months == 0:
+                    self.handle_undelivered(identity_uuid)
 
 
 process_whatsapp_timeout_system_event = ProcessWhatsAppTimeoutSystemEvent()
--- a/changes/tasks.py
+++ b/changes/tasks.py
@@ -1522,7 +1522,7 @@ class ProcessWhatsAppTimeoutSystemEvent(Task):
             if timeout_timestamp is None:
                 self.handle_undelivered(identity_uuid)
             else:
-                d1 = datetime.strptime(timeout_timestamp, "%d/%m/%Y")
+                d1 = datetime.fromtimestamp(timeout_timestamp)
                 d2 = datetime.today()
                 week1 = d1 - timedelta(days=d1.weekday())
                 week2 = d2 - timedelta(days=d2.weekday())
--- a/changes/tasks.py
+++ b/changes/tasks.py
@@ -1,5 +1,6 @@
 import json
 import re
+import time
 from datetime import datetime, timedelta
 from itertools import chain as ichain
 from itertools import dropwhile, takewhile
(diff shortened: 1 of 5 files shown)


changes/views.py
```diff
@@ -287,6 +287,24 @@ def post(self, request, *args, **kwargs):
...
+        if not serializer.is_valid():
+            return Response(status=status.HTTP_400_BAD_REQUEST)
+
+        for item in serializer.validated_data["statuses"]:
+            for error in serializer.validated_data["errors"]:
+                tasks.process_whatsapp_timeout_system_event.delay(
+                    item["id"], item["timestamp"], item["recipient_id"], item["errors"]
+                )
+
+        return Response(status=status.HTTP_202_ACCEPTED)
```
rudigiesler: I think we rather want to expand the existing view https://github.com/praekeltfoundation/ndoh-hub/blob/ee891d316964398f834b725ec84242d212dc6cdc/changes/views.py#L239-L272 instead of making a new view to handle this. The events are already going to that old view, it's just that nothing is being done with them

changes/tasks.py
```diff
@@ -1468,6 +1469,65 @@ def run(self, vumi_message_id: str, event_type: str, **kwargs) -> None:
...
+                ]
+            )["to_identity"]
+        except StopIteration:
+            """
+            Outbound with message id doesn't exist, so don't continue
+            """
+            return
+
+        if event_type == "undelivered":
+            d1 = datetime.fromtimestamp(timestamp)
```
rudigiesler: This timestamp is when the event was created, not when we last sent them the SMS.

The logic should be similar to:
 - We receive this kind of event for the first time, we send them an SMS.
 - If we receive the event a second time, then we need to check when we sent that last SMS. If we sent it less than a month ago, then we do nothing, if it's been more than a month, then we send them that same SMS again.

This means that we need to store when we sent that last SMS somewhere, so that we can look it up when we get another event, and determine if we should send another SMS or not. A good place to store this is probably on the identity, as a field in the identity's `details`

changes/test_tasks.py
```diff
@@ -311,6 +312,30 @@ def test_no_message_found(self, mock_create_outbound):
...
+                "to_identity": "test-identity-uuid",
+                "content": (
+                    "We see that your MomConnect WhatsApp messages are not being "
+                    "delivered. If you would like to receive your messages over "
+                    "SMS, reply ‘SMS’."
+                ),
+                "channel": "JUNE_TEXT",
+                "metadata": {},
+            }
+        )
```
rudigiesler, line 376: We probably want at least two tests here, one that sends an SMS message, and one that doesn't

changes/test_tasks.py
```diff
@@ -311,6 +312,30 @@ def test_no_message_found(self, mock_create_outbound):
...
+    @mock.patch("changes.tasks.utils.ms_client.create_outbound")
+    @responses.activate
+    def test_timeout_message_sent_delivered(self, mock_create_outbound):
+        """
+        The task should send the correct outbound based on the delivered event.
+        """
+        self.create_outbound_lookup()
+        self.create_identity_lookup()
+
+        process_whatsapp_timeout_system_event("messageid", 1542844800, "undelivered")
```
rudigiesler: The issue with hard-coding the timestamp here is that, although this might work now, `datetime.now()` is always going to increase, so as time passes you'll get a bigger value for `datetime.now()`, but this value will stay the same, so eventually this test will fail, even though no code has changed.

This should rather be a value that is relative to `datetime.now()`

changes/test_tasks.py
```diff
@@ -311,6 +312,30 @@ def test_no_message_found(self, mock_create_outbound):
         )
         self.assertFalse(mock_create_outbound.called)
 
+    @mock.patch("changes.tasks.utils.ms_client.create_outbound")
+    @responses.activate
+    def test_timeout_message_sent_delivered(self, mock_create_outbound):
+        """
+        The task should send the correct outbound based on the delivered event.
+        """
+        self.create_outbound_lookup()
```
rudigiesler, line 354: We don't need to create the outbound lookup here, since we're mocking it out with the `@mock.patch`

OMosimege: when I remove that line, I get a `requests.exceptions.ConnectionError: Connection refused: GET http://ms/api/v1/outbound/?vumi_message_id=messageid` error

changes/test_tasks.py
OMosimege: thanks for pointing this out

changes/test_tasks.py
OMosimege: will do

changes/tasks.py
OMosimege: makes sense, thank you

changes/views.py
OMosimege: when we last talked, you said to create a new view, and if the new one passes, we delete the old one

changes/test_tasks.py
rudigiesler: Oh right, sorry I got confused. The thing we're mocking out is the creation of a new message, and this adds the lookup for an existing message, they're two different things. We should then leave this here.

changes/views.py
rudigiesler: Sorry, I think we misunderstood each other. I meant that we add this functionality to the existing `ReceiveWhatsAppEvent` view, and then we can delete the old `ReceiveWhatsAppSystemEvent` view

changes/tasks.py
```diff
@@ -1468,6 +1468,88 @@ def run(self, vumi_message_id: str, event_type: str, **kwargs) -> None:
...
+                    )
+
+                    utils.ms_client.create_outbound(
+                        {
+                            "to_identity": identity_uuid,
+                            "content": text,
+                            "channel": "JUNE_TEXT",
+                            "metadata": {},
+                        }
+                    )
```
rudigiesler: I think we can be a bit more DRY here, and extract the sending of the message into a function, so that we don't have to repeat the exact code twice.

changes/tasks.py
```diff
@@ -1468,6 +1468,88 @@ def run(self, vumi_message_id: str, event_type: str, **kwargs) -> None:
...
+                )
+        else:
+            d1 = datetime.fromtimestamp(details["timeout_timestamp"])
+            d2 = datetime.today()
+            week1 = d1 - timedelta(days=d1.weekday())
+            week2 = d2 - timedelta(days=d2.weekday())
+            # Returns 0 if both dates fall withing one week, 1 if on two weeks etc.
+            weeks = int(round((week2 - week1).days / 7))
+            months = int(round((week2 - week1).days / 30))
+            if weeks == 1 and months == 0:
```
rudigiesler: We want this to be a month by default, but also it would be nice for this to be configurable.

This logic can also be simplified quite a bit. When you subtract two datetime objects, you get a `timedelta` object of the difference between the two. https://docs.python.org/3/library/datetime.html#timedelta-objects

We can also be using `datetime.now` instead of `datetime.today` to get a more accurate time difference, but we should also probably also ensure that we're using timezone aware timestamps, so that the time comparison is more accurate.

So we could do something like:
```
if (d2 - d1).days >= settings.WHATSAPP_EXPIRY_SMS_BOUNCE_DAYS:
    ....
```

changes/tasks.py
```diff
@@ -1468,6 +1468,88 @@ def run(self, vumi_message_id: str, event_type: str, **kwargs) -> None:
...
+    message informing them and give the option to switch to SMS.
+    """
+
+    name = "ndoh_hub.changes.tasks.process_whatsapp_timeout_system_event"
+
+    def handle_undelivered(self, user_id, source_id, identity_uuid):
+        identity = is_client.get_identity(identity_uuid)
+        details = identity["details"]
+        if "timeout_timestamp" not in details:
+            identity["details"]["timeout_timestamp"] = datetime.today()
```
rudigiesler: Just setting it here doesn't store it in the identity store, we need to call the `is_client.update_identity` to ensure that this value is stored.

changes/tasks.py
OMosimege: Okay will do

changes/tasks.py
OMosimege: am I able to add 2 configurable dates in `WHATSAPP_EXPIRY_SMS_BOUNCE_DAYS` in settings, adding that it should be a week, but if it is within 7+ days but less than a month it shouldn't send?

changes/tasks.py
```diff
@@ -1468,6 +1469,78 @@ def run(self, vumi_message_id: str, event_type: str, **kwargs) -> None:
...
+                        "metadata": {},
+                    }
+                )
+
+    def handle_undelivered(self, user_id, source_id, identity_uuid):
+        identity = is_client.get_identity(identity_uuid)
+        for item in identity["results"]:
+            details = item["details"]
+            if "timeout_timestamp" not in details:
+                details["timeout_timestamp"] = time.mktime(datetime.now().timetuple())
```
rudigiesler: Can also do `datetime.now().timestamp()`, which is a little simpler. We should also probably be using `datetime.utcnow()`, to ensure that the timestamp we're storing is in UTC, and not whatever local time the server happens to have.

changes/tasks.py
```diff
@@ -1468,6 +1469,78 @@ def run(self, vumi_message_id: str, event_type: str, **kwargs) -> None:
...
+                is_client.update_identity({"id": item["id"], "details": details})
+                self.send_outbound(user_id, source_id, identity_uuid)
+
+            else:
+                d1 = datetime.fromtimestamp(details["timeout_timestamp"])
+                d2 = datetime.now()
```
rudigiesler: We should be using `datetime.utcnow()` here, especially if we make the change to store the timestamp as utc above, so that the two times we're comparing are both UTC times, and not the local timezone of the server.

changes/test_tasks.py
```diff
@@ -80,6 +83,35 @@ def create_outbound_lookup(self, result_count=1):
...
 
+    def create_identity_lookup_with_id_field(self, lang="eng_ZA"):
+        responses.add(
+            responses.GET,
+            "http://is/api/v1/identities/test-identity-uuid/",
+            json={
+                "results": [
+                    {"id": "test_identity-uuid", "details": {"lang_code": lang}}
+                ]
+            },
```
rudigiesler: This is not the correct return value for an identity lookup. Only when we ask for a list of identities (`GET /api/v1/identities/`, will we get a `results` with a list of identities. If we ask for a specific identity (`GET /api/v1/identities/<identity-uuid>/`, then we will just get the details of that identity, not a list of results.

changes/test_tasks.py
```diff
@@ -311,6 +362,121 @@ def test_no_message_found(self, mock_create_outbound):
         )
         self.assertFalse(mock_create_outbound.called)
 
+    @mock.patch("changes.tasks.utils.ms_client.create_outbound")
+    @mock.patch("changes.tasks.is_client.update_identity")
+    @responses.activate
+    def test_timeout_error(self, mock_update_identiity, mock_create_outbound):
+        """
+        The task should create a Change according to the details received from the
+        message sender
```
rudigiesler: This docstring is incorrect. Should probably say something about sending an outbound message

changes/test_tasks.py
```diff
@@ -311,6 +362,121 @@ def test_no_message_found(self, mock_create_outbound):
...
+
+    @mock.patch("changes.tasks.utils.ms_client.create_outbound")
+    @mock.patch("changes.tasks.is_client.update_identity")
+    @responses.activate
+    def test_timeout_error_with_timestamp(
+        self, mock_update_identiity, mock_create_outbound
+    ):
+        """
+        The task should create a Change according to the details received from the
+        message sender
```
rudigiesler: Same here with this docstring

changes/test_tasks.py
```diff
@@ -311,6 +362,121 @@ def test_no_message_found(self, mock_create_outbound):
...
+    @mock.patch("changes.tasks.utils.ms_client.create_outbound")
+    @mock.patch("changes.tasks.is_client.update_identity")
+    @responses.activate
+    def test_timeout_error_no_outbound_send(
+        self, mock_update_identiity, mock_create_outbound
+    ):
+        """
+        The task should create a Change according to the details received from the
+        message sender
+        """
```
rudigiesler, line 440: And this docstring

changes/tasks.py
OMosimege: using `.timetuple()` is easier that `.timestamp(), because it rounds of the milliseconds of the timestamp i.e.  a .timestamp() will create a timestamp of 1543932583.366118 and 1543932583.370697 within seconds, while .tuple() will create 1543932483.0. Should I still change it?

changes/tasks.py
OMosimege: thank you. will implement that

changes/test_tasks.py
OMosimege: thanks for picking this up

changes/tasks.py
rudigiesler: Yes I don't think that there's any harm in having more accuracy. It's not super important for this case, but it does make the code a bit easier to read, which is important

changes/test_tasks.py
OMosimege: thanks

changes/test_tasks.py
OMosimege: thanks

changes/test_tasks.py
OMosimege: I am unable to create a mock assert ```{
                "test_identity-uuid",
                "details": {"lang_code": "eng_ZA", "timeout_timestamp": timestamp},
            }```

I get a python error afterwards, how do I achieve this?

changes/tasks.py
rudigiesler: No, I think we only need a single days integer, eg let's say its value is 30.

If the last message was sent less than 30 days ago, then we ignore the event and don't send anything.

If the last message was sent more than 30 days ago, then we send the message again.

changes/test_tasks.py
rudigiesler: I think then in your code, there's something that's expecting a `results` key with a list of identities. It's then correct that the test is failing, because the actual API won't return that, so if we ran it on the real identity store, we would get the same python error. So we should fix the code that's processing the result of the call to the identity store

changes/tasks.py
```diff
@@ -1468,6 +1468,82 @@ def run(self, vumi_message_id: str, event_type: str, **kwargs) -> None:
...
+
+    name = "ndoh_hub.changes.tasks.process_whatsapp_timeout_system_event"
+
+    def return_date_time(self):
+        return datetime.utcnow()
+
+    def send_outbound(self, user_id, source_id, identity_uuid):
+        identity = is_client.get_identity(identity_uuid)
+        # Transform to django language code
+        for item in identity["results"]:
```
rudigiesler: Since we're using `get_identity`, this won't have a `results` key with a list of identities. This will instead just be that single identity that we're looking for

changes/tasks.py
```diff
@@ -1468,6 +1468,82 @@ def run(self, vumi_message_id: str, event_type: str, **kwargs) -> None:
...
+                        "to_identity": identity_uuid,
+                        "content": text,
+                        "channel": "JUNE_TEXT",
+                        "metadata": {},
+                    }
+                )
+
+    def handle_undelivered(self, user_id, source_id, identity_uuid):
+        identity = is_client.get_identity(identity_uuid)
+        for item in identity["results"]:
```
rudigiesler: Same here with list of identities vs single identity

changes/tasks.py
```diff
@@ -1468,6 +1468,82 @@ def run(self, vumi_message_id: str, event_type: str, **kwargs) -> None:
...
+            details = item["details"]
+            if "timeout_timestamp" not in details:
+                date = self.return_date_time()
+                details["timeout_timestamp"] = date.timestamp()
+                is_client.update_identity({"id": item["id"], "details": details})
```
rudigiesler: `update_identity` takes two arguments, the ID of the identity, and the object to use to update. I think we rather want this to look like `is_client.update_identity(item["id"], {"details": details})`

changes/tasks.py
```diff
@@ -1468,6 +1468,82 @@ def run(self, vumi_message_id: str, event_type: str, **kwargs) -> None:
...
+                self.send_outbound(user_id, source_id, identity_uuid)
+
+            else:
+                d1 = datetime.fromtimestamp(details["timeout_timestamp"])
+                d2 = self.return_date_time()
+                delta = (d2 - d1).days
+                if delta >= settings.WHATSAPP_EXPIRY_SMS_BOUNCE_DAYS:
+                    details["timeout_timestamp"] = d2.timestamp()
+                    is_client.update_identity({"id": item["id"], "details": details})
+                    self.send_outbound(user_id, source_id, identity_uuid)
```
rudigiesler: Maybe we can just send the whole identity here, so that we don't have to do the second HTTP request in `send_outbound`

changes/test_tasks.py
```diff
@@ -80,6 +82,35 @@ def create_outbound_lookup(self, result_count=1):
...
+        responses.add(
+            responses.PATCH,
+            "http://is/api/v1/identities/test-identity-uuid/",
+            json={
+                "results": [
+                    {
+                        "id": "test_identity-uuid",
+                        "details": {"lang_code": lang, "timeout_timestamp": "time"},
+                    }
+                ]
```
rudigiesler: The `PATCH` action on this endpoint will just return the updated identity, not a list of identities.

changes/test_tasks.py
```diff
@@ -89,6 +120,25 @@ def create_identity_lookup(self, lang="eng_ZA"):
...
+            json={
+                "results": [
+                    {
+                        "id": "test_identity-uuid",
+                        "details": {
+                            "lang_code": "eng_ZA",
+                            "timeout_timestamp": timestamp,
+                        },
+                    }
+                ]
```
rudigiesler: This endpoint doesn't return a list of results, rather, it just returns the specific identity

changes/tasks.py
OMosimege: ahh yes thank you

changes/tasks.py
OMosimege: yes

changes/tasks.py
OMosimege: By send the whole identity, do you mean send it in the update_identityline?  is the a method to do both an update and a send

rudigiesler: I meant that instead of doing a `self.send_outbound(user_id, source_id, identity_uuid)`, and then performing another request to the identity store in `send_outbound`, we should rather do a `self.send_outbound(user_id, source_id, identity)`, and then the `send_outbound` function has all the identity details, and doesn't need to do a request to the identity store.

OMosimege: oh I see what you meant, that's brilliant thank you

rudigiesler (approved):
👍 🎆 🍾 🍰

PR #191, opened 2018-12-03
Fix bugs in submission of helpdesk responses to DHIS2

rudigiesler:
This PR addresses two issues that we've been having with this task

1. It's possible that the type of the message is `null`, we need to handle that and fallback to manually trying to get the content for the content types that we know about
2. It's possible that a message has a timestamp of `null`, in which case we need to fallback to the `inserted_at` timestamp


OMosimege (approved):
Everything looks good 🍰

OMosimege (approved):
Everything looks good 🍰

PR #193, opened 2018-12-12
Add Prometheus metrics

rudigiesler:
This PR adds default Prometheus metrics scraping for python and Django, and a restricted metrics endpoint for exported the metrics.

It also removes any metric that are no longer required.


OMosimege (approved):
Looks good 🍰

PR #194, opened 2018-12-12
Fix internal_only decorator

rudigiesler:
The nginx inside the docker container adds a value to the X-Forwarded-For header, so if something is an internal request it has 1 not 0 values in the header, and if something goes through our load balancer it has 2 not 1 values in the header.

This PR also adds a docker-compose setup, to make testing these sorts of things easier.


OMosimege (approved):
Looks good 👍

PR #197, opened 2019-01-09
Add handling for additional message types from Turn

rudigiesler:
This PR ensures that we can handle some additional message types, so that the
submission to openhim doesn't fail.


OMosimege (approved):
Looks good 🍰

PR #198, opened 2019-01-10
Add retries for HTTP failures for process_whatsapp_timeout_system_event task

rudigiesler:
This PR splits the task up into separate idempotent tasks, and then adds a
retry strategy to each task to retry in the event of an HTTP error


OMosimege (approved):
Looks great 👍

PR #199, opened 2019-01-14
Feature/momza 1111 fix loss switch lang code

OMosimege:
Currently for loss switches, we look for the language code in lang_code on the identity, but it seems like not all identities have this field. 

This PR fixes by setting the language the same as their existing messageset


rudigiesler (approved):
Looks great 🍰 👍

PR #200, opened 2019-01-30
added endpoint for identity with missing msisdn

OMosimege:
Currently, if there is no to_addr, we get an error later on when we try to send the message. We should rather ensure that to_addr is not None earlier on, before we try to send that message, and handle the issue there.

In this PR we are creating an endpoint for and webhook fired for a failed msisdn lookup


changes/urls.py
```diff
@@ -12,6 +12,9 @@
 urlpatterns = [
     url(r"^api/v1/", include(router.urls)),
     url(r"^api/v1/change/inactive/$", views.OptOutInactiveIdentity.as_view()),
+    url(
+        r"^api/v1/change/inactive/$", views.SeedMessageSenderFailedMsisdnHook.as_view()
+    ),
```
rudigiesler, line 19: I don't think this is what we want. Because it's the same regex as the previous one, this view will never be called.

I think we should have a unit test that calls this URL, that would pick up an issue like this

changes/views.py
```diff
@@ -310,3 +311,32 @@ def post(self, request):
...
         tasks.process_whatsapp_contact_check_fail.delay(str(request.user.pk), msisdn)
         return Response(status=status.HTTP_202_ACCEPTED)
+
+
+class SeedMessageSenderFailedMsisdnHook(generics.GenericAPIView):
+    """
+    Receives events from the Seed Message Sender. Supports:
+    * identity.no_address hook
+
+    Requires token auth in the querystring "token" field.
```
rudigiesler, line 321: This line isn't true, because there's no authentication on this view. I think we should have a unit test to ensure that authentication is required, because we don't want an unauthenticated endpoint on our API

changes/views.py
```diff
@@ -310,3 +311,32 @@ def post(self, request):
...
+
+    def post(self, request, *args, **kwargs):
+        try:
+            # The hooks send the request data as {"hook":{}, "data":{}}
+            data = request.data["data"]
+        except KeyError:
+            raise ValidationError('"data" must be supplied')
+        identity_id = data.get("identity_id", None)
+        if identity_id is None or identity_id == "":
+            raise ValidationError('"identity_id" must be supplied')
```
rudigiesler: It doesn't seem like we're using the `serializer_class` anywhere.

I think we should rather be using the serializer to validate the request, rather than manually checking that the required fields are present.

changes/serializers.py
```diff
@@ -127,3 +127,17 @@ class Data(serializers.Serializer):
...
+class SeedMessageSenderFailedMsisdnHookSerializer(serializers.Serializer):
+    class Hook(serializers.Serializer):
+        id = serializers.IntegerField()
+        event = serializers.ChoiceField(choices=["identity.no_address"])
+        target = serializers.CharField()
+
+    hook = Hook()
+
+    class Data(serializers.Serializer):
+        address = PhoneNumberField(country_code="ZA")
```
rudigiesler: I don't think that this is the correct format for this webhook. Looking at the message sender: https://github.com/praekeltfoundation/seed-message-sender/blob/develop/message_sender/tasks.py#L218-L223 , it seems like the `data` field has a `to_identity` field, that is a UUID, not an `address` field, that has a phone number.

OMosimege: correct, thank you

changes/views.py
OMosimege: Thanks for pointing this out

changes/views.py
OMosimege: Okay will remove that part

changes/test_views.py
```diff
@@ -426,3 +426,40 @@ def test_whatsapp_contact_check_failure(self, task):
...
+                    "id": 1,
+                    "event": "identity.no_address",
+                    "target": "http://example.org",
+                },
+                "data": {"to_identity": "cbaf27bc-2ba9-4e4a-84c1-098d5abd80bf"},
+            },
+            format="json",
+        )
+
+        self.assertEqual(response.status_code, status.HTTP_202_ACCEPTED)
```
rudigiesler, line 509: I think we need to assert a bit more than a 202 here. We probably want to assert that the `Change` was created with the correct data

changes/views.py
```diff
@@ -310,3 +311,20 @@ def post(self, request):
...
+
+    Requires token auth in the querystring "token" field.
+    """
+
+    serializer_class = SeedMessageSenderFailedMsisdnHookSerializer
+    authentication_classes = (TokenAuthQueryString,)
+
+    def post(self, request, *args, **kwargs):
+        serializer = self.get_serializer(data=request.data)
+        serializer.is_valid(raise_exception=True)
```
rudigiesler, line 329: We need to create the `Change` object here, otherwise this view doesn't do the action, it just validates the request and then returns

changes/serializers.py
```diff
@@ -127,3 +129,17 @@ class Data(serializers.Serializer):
...
+class SeedMessageSenderFailedMsisdnHookSerializer(serializers.Serializer):
+    class Hook(serializers.Serializer):
+        id = serializers.IntegerField()
+        event = serializers.ChoiceField(choices=["identity.no_address"])
+        target = serializers.CharField()
+
+    hook = Hook()
+
+    class Data(serializers.Serializer):
+        identity = str(uuid.uuid4())
```
rudigiesler: I don't think that this will validate that `identity` is a UUID, this needs to be an instance of the UUIDField: https://www.django-rest-framework.org/api-guide/fields/#uuidfield

changes/test_views.py
```diff
@@ -426,3 +432,84 @@ def test_whatsapp_contact_check_failure(self, task):
...
+        self.identity_lookup(registrant_id)
+        self.jembi_post()
+
+        change_data = {
+            "registrant_id": "cbaf27bc-2ba9-4e4a-84c1-098d5abd80bf",
+            "action": "momconnect_nonloss_optout",
+            "data": {"reason": "missing_to_addr"},
+            "source": source,
+        }
+        change = Change.objects.create(**change_data)
```
rudigiesler: We don't want to be creating the Change here in the test. The view should be creating the `Change` for us, so the `self.client.post` should create the `Change`. What we should be doing here is fetching the `Change` from the database, and asserting that the fields were created correctly.

changes/serializers.py
```diff
@@ -127,3 +127,17 @@ class Data(serializers.Serializer):
...
+class SeedMessageSenderFailedMsisdnHookSerializer(serializers.Serializer):
+    class Hook(serializers.Serializer):
+        id = serializers.IntegerField()
+        event = serializers.ChoiceField(choices=["identity.no_address"])
+        target = serializers.CharField()
+
+    hook = Hook()
+
+    class Data(serializers.Serializer):
+        registrant_id = serializers.UUIDField()
```
rudigiesler: In the webhook: https://github.com/praekeltfoundation/seed-message-sender/blob/develop/message_sender/tasks.py#L218-L223 , this field is named `to_identity`, not `registrant_id`.

changes/test_views.py
```diff
@@ -426,3 +432,84 @@ def test_whatsapp_contact_check_failure(self, task):
...
+
+        response = self.client.post(
+            url,
+            data={
+                "hook": {
+                    "id": 1,
+                    "event": "identity.no_address",
+                    "target": "http://example.org",
+                },
+                "data": {"registrant_id": registrant_id},
```
rudigiesler: Looking at the webhook: https://github.com/praekeltfoundation/seed-message-sender/blob/develop/message_sender/tasks.py#L218-L223 , this field should be named `to_identity` not `registrant_id`

changes/test_views.py
OMosimege: how do we fetch the change from the database?

rudigiesler (approved):

PR #203, opened 2019-02-06
Add setting for toggling unsent event action

rudigiesler:
If WhatsApp disable our template, then they send us the same event as when a
user's phone doesn't support tempalted messages, so we need to have a way to
switch this feature off in case our template gets disabled


OMosimege (approved):
Great work

PR #205, opened 2019-02-21
Change to using new whatsapp templates

rudigiesler:
We can no longer use the original blank template that we had, so we now need to
use new, more specific templates when sending whatsapp templated messages


OMosimege (approved):
Looks good 👍

rudigiesler:
Won't be merging this in yet, waiting first for the templates to be submitted and approved

PR #206, opened 2019-03-14
Feature/momza 1155 migrate existing messagesets to use new templates

OMosimege:
This PR adds scripts for adding WhatsApp template metadata to messagesets


changes/tasks.py
```diff
@@ -534,7 +534,43 @@ def switch_channel(self, change):
...
+            reason = change.data and change.data.get("reason")
+            if reason != "whatsapp_unsent_event":
+                utils.ms_client.create_outbound(
+                    {
+                        "to_identity": change.registrant_id,
+                        "content": text,
+                        "channel": "WHATSAPP",
+                        "metadata": {
+                            "template": {
+                                "name": "important_info",
```
OMosimege: does this template name need to change?

changes/tests.py
```diff
@@ -3868,10 +3869,23 @@ def test_switch_channel_to_sms_not_available(self, mock_create_outbound):
...
                     "between 1 - 2 are only available on WhatsApp - switching to "
                     "SMS means you will not receive any messages. You can stop "
-                    "your MomConnect messages completely by replying ‘STOP‘"
+                    "your MomConnect messages completely by replying 'STOP'"
                 ),
                 "channel": "WHATSAPP",
-                "metadata": {},
+                "metadata": {
+                    "template": {
+                        "name": "important_info",
```
OMosimege: same here

changes/tasks.py
rudigiesler: Ah, yes it does, but I think that should happen on the other PR (since that's where this change is from)

PR #208, opened 2019-03-21
Update EngageActionView to get option and return refresh header

smn:
The Turn API for this has changed slightly. The `option` is no longer in
the `payload` but a top level attribute instead.

Also, we can force a refresh of the UI by returning the `X-Turn-Integration-Refresh` header
set to `true`.


smn:
@erikh360 & @rudigiesler I've updated the context action view for the changes I've made in Turn.

erikh360 (approved):
Looks good!

PR #210, opened 2019-04-03
Update language mapping

rudigiesler:
WhatsApp don't like our solution of mapping unsupported languages to supported
ones, so we need to make all non-supported languages point to english templates


OMosimege (approved):
looks good 👍

PR #212, opened 2019-04-18
Refresh context on completion of action

rudigiesler:
Currently, we're using the X-Turn-Integration-Refresh header, which immediately refreshes the context after sending the action through to us.

The issue with this is that the action hasn't completed on our side yet, so when it refreshes it still displays the old data.

What we should instead be doing is using the "Delayed refresh" function of the API, where we call the API back with a webhook once we have finished processing the action, so that the context can be updated with the changed details

### DIFF
--- a/changes/models.py
+++ b/changes/models.py
@@ -73,5 +73,17 @@ class Change(models.Model):
     )
     user = property(lambda self: self.created_by)
 
+    @property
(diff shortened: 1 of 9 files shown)


KaitCrawford (approved):
:+1:

PR #213, opened 2019-05-02
Feature/feature/momza 1096 adding missing translations

OMosimege:
In this PR, we add missing translations for postbirth Auto Responses for Moms switching SMS/WA

### DIFF
--- a/changes/tasks.py
+++ b/changes/tasks.py
@@ -570,16 +570,7 @@ class ValidateImplement(Task):
                     {
                         "to_identity": change.registrant_id,
                         "content": fail_contact_check_text,
-                        "channel": "sms",
-                        "metadata": {
-                            "template": {
-                                "name": "mc_important_info",
-                                "language": utils.WHATSAPP_LANGUAGE_MAP[
-                                    subscriptions[0]["lang"]
-                                ],
-                                "variables": [fail_contact_check_text],
-                            }
-                        },
+                        "channel": "JUNE_TEXT",
                     }
                 )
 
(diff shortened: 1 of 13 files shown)


changes/tasks.py
```diff
@@ -548,8 +554,54 @@ def switch_channel(self, change):
...
+                    "MomConnect msgs for kids aged 1-2 are only on WA. "
+                    "To stop msgs, reply 'STOP' (std rates apply)"
+                )
+
+            if reason == "whatsapp_contact_check_fail":
+                utils.ms_client.create_outbound(
+                    {
+                        "to_identity": change.registrant_id,
+                        "content": fail_contact_check_text,
+                        "channel": "WHATSAPP",
```
rudigiesler: If it's a whatsapp contact check fail, that means that they're not on WhatsApp, so we can't send this message over WhatsApp, so we probably want to send this message over SMS. The same with the unsent event.

OMosimege: thanks for picking this up, I missed it

changes/tasks.py
```diff
@@ -548,8 +554,54 @@ def switch_channel(self, change):
...
+                        },
+                    }
+                )
+
+            elif reason == "postbirth_wa_unsent_event":
+                utils.ms_client.create_outbound(
+                    {
+                        "to_identity": change.registrant_id,
+                        "content": unsent_event_text,
+                        "channel": "sms",
```
rudigiesler: I think that the channel name here is `JUNE_TEXT`.

Also, we won't need the template metadata for SMS.

OMosimege: ahh I see, thank you

rudigiesler (approved):

PR #214, opened 2019-05-07
add new view for jembi health check

miltontony:


registrations/views.py
```diff
@@ -363,6 +363,64 @@ def get(self, request, *args, **kwargs):
...
+
+            jembi_result = result.json()
+
+            if len(jembi_result.get("rows")) > 0:
+                resp = {
+                    "up": True,
+                    "result": {"Facility": jembi_result.get("rows")[0][2]},
+                }
+            else:
+                raise Http404()
```
miltontony: @rudigiesler just a check on this - should we return 404 for when the service is down? or is there a better error code?

rudigiesler: @miltontony I think there are 2 codes here depending on what the failure is.
502 - we received a bad response from upstream
504 - we timed out requesting from upstream

registrations/views.py
```diff
@@ -363,6 +363,64 @@ def get(self, request, *args, **kwargs):
...
+                    "400 Error when posting to Jembi.\n"
+                    "Response: %s\nPayload:%s" % (e.response.text)
+                )
+                return Response(
+                    "Error when posting to Jembi. Body: %s Payload: %r"
+                    % (e.response.content),
+                    status=status.HTTP_400_BAD_REQUEST,
+                )
+            else:
+                raise e
```
miltontony, line 415: @rudigiesler that will be handled here -- the HTTP error will just be passed through to the monitor.

registrations/views.py
```diff
@@ -363,6 +363,64 @@ def get(self, request, *args, **kwargs):
...
+            )
+
+        try:
+            clinic_code = request.query_params.get("clinic_code")
+            result = requests.get(
+                urljoin(settings.JEMBI_BASE_URL, "NCfacilityCheck"),
+                headers={"Content-Type": "application/json"},
+                auth=(settings.JEMBI_USERNAME, settings.JEMBI_PASSWORD),
+                params={"criteria": "value:{}".format(clinic_code)},
+                verify=False,
```
rudigiesler, line 393: Since this is in the view, I think we should specify a timeout, we don't want this tying up a worker for a very long period of time.

registrations/views.py
rudigiesler, line 415: We should probably handle the exception and return a response here. We don't want sentry alerts because their API is down, it's not this service that's broken.

registrations/views.py
OMosimege: @rudigiesler how do we add a timeout?

OMosimege: okay thanks

registrations/views.py
```diff
@@ -363,6 +363,66 @@ def get(self, request, *args, **kwargs):
...
+                timeout=300.0,
```
rudigiesler: I think 5 minutes is a bit too long? Was thinking more like 10 seconds.

registrations/views.py
rudigiesler: I think the issue here is that because we're raising the exception here, we'll get a Sentry error

registrations/views.py
rudigiesler: Looking at this again, I think a 502 error is more suited here

registrations/views.py
```diff
@@ -363,6 +363,66 @@ def get(self, request, *args, **kwargs):
...
+                    "Response: %s\nPayload:%s" % (e.response.text)
+                )
+                return Response(
+                    "Error when posting to Jembi. Body: %s Payload: %r"
+                    % (e.response.content),
+                    status=status.HTTP_400_BAD_REQUEST,
+                )
+            else:
+                raise e
+                return Response("Timeout", status.HTTP_504_GATEWAY_TIMEOUT)
```
rudigiesler: I don't think that this line will ever be reached?

OMosimege: both line 420 and 421?

rudigiesler: Just 421. Line 420 will reraise the exception, which means that we'll never be able to get to line 421

OMosimege: in your opinion, if I remove line 421 and replace it on line 406, (where the timeout of the request will be), would that make more sense?

rudigiesler: We probably want to catch the timeout error separately, and return the timeout response in that case. Line 406 means that we've gotten a successful response, so it hasn't timed out.

registrations/views.py
```diff
@@ -363,6 +363,72 @@ def get(self, request, *args, **kwargs):
...
+                return Response(
+                    "Error when posting to Jembi. Body: %s Payload: %r"
+                    % (e.response.content),
+                    status=status.HTTP_400_BAD_REQUEST,
+                )
+            else:
+                raise e
+
+        except (requests.exceptions.Timeout,) as e:
+            if e.response.status_code == 504:
```
rudigiesler: If we have a timeout, then we won't have a response (a timeout means that the server hasn't sent us a response in time)

OMosimege: so, I should remove the response from the exception?

rudigiesler: I think we should just remove the `if`. If we get a `Timeout` exception, then we should just return a 504.

OMosimege: I'm concerned that I do not know how to effectively test out and mock a 504 error

rudigiesler: Yes, that is something that I don't think we can really mock out. We don't want the tests waiting around for 10 seconds for a timeout to happen. I think for this case, it's more work to be able to create a unit test for it than it's worth, so I think we can leave that out for now.

rudigiesler (approved):

PR #218, opened 2019-05-28
Active subscriptions API

rudigiesler:
This PR adds an API endpoint for querying if a user has active subscriptions, and if they do, what kind of subscription they have (public, clinic, or none)


OMosimege (approved):
Great work 👍

PR #219, opened 2019-05-29
Add optout status to subscription check view

rudigiesler:
We need to know if the user is currently opted out to know whether to ask them to opt in again or not.

We're adding this to the subscription check view to reduce the amount of HTTP requests needed during registration, since all the information we need is available in this view already.


OMosimege (approved):
Good work 🍰

PR #221, opened 2019-06-04
Feature/momza 1218 add api for facility code check

OMosimege:
In this PR we check if the facility code is valid, so we should create a flow hook endpoint for Rapidpro to be able to check if the clinic code is valid.

### DIFF
--- a/registrations/test_views.py
+++ b/registrations/test_views.py
@@ -326,9 +326,7 @@ class FacilityCodeCheckViewTests(AuthenticatedAPITestCase):
         )
         response = self.normalclient.get(url)
         self.assertEqual(response.status_code, 200)
-        self.assertEqual(
-            json.loads(response.content)["result"]["Facility"], "invalid"
-        )
+        self.assertEqual(json.loads(response.content)["result"]["Facility"], "invalid")
 
 
 class PositionTrackerViewsetTests(AuthenticatedAPITestCase):
--- a/registrations/test_views.py
+++ b/registrations/test_views.py
@@ -306,7 +306,7 @@ class FacilityCodeCheckViewTests(AuthenticatedAPITestCase):
         """
             Test on Facility Code Check when Jembi return empty array for
             wrong code given
-            GET - returns 404 response
+            GET - returns 200 response
         """
 
         clinic_code = 111111
@@ -325,7 +325,7 @@ class FacilityCodeCheckViewTests(AuthenticatedAPITestCase):
             reverse("facilitycode-check"), urlencode({"clinic_code": clinic_code})
         )
         response = self.normalclient.get(url)
-        self.assertEqual(response.status_code, 404)
+        self.assertEqual(response.status_code, 200)
 
 
 class PositionTrackerViewsetTests(AuthenticatedAPITestCase):
--- a/registrations/test_views.py
+++ b/registrations/test_views.py
@@ -268,7 +268,9 @@ class FacilityCodeCheckViewTests(AuthenticatedAPITestCase):
         )
         response = self.normalclient.get(url)
         self.assertEqual(response.status_code, 200)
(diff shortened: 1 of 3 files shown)


registrations/views.py
```diff
@@ -1095,6 +1095,57 @@ def prune(self, request):
...
+                auth=(settings.JEMBI_USERNAME, settings.JEMBI_PASSWORD),
+                params={"criteria": "value:{}".format(clinic_code)},
+                verify=False,
+            )
+            result.raise_for_status()
+            jembi_result = result.json()
+            print(jembi_result)
+            resp = {"Facility": jembi_result.get("rows")[0][2]}
+            print("views.resp")
+            print(resp)
```
rudigiesler: We should take these print statements out

registrations/test_views.py
```diff
@@ -235,6 +235,69 @@ def test_get_registration_only_by_user(self):
...
+                urlencode({"criteria": "value:{}".format(clinic_code)})
+            ),
+            json=result,
+            status=400,
+        )
+        url = "{}?{}".format(
+            reverse("facilitycode-check"), urlencode({"clinic_code": clinic_code})
+        )
+        response = self.normalclient.get(url)
+        self.assertEqual(response.status_code, 400)
```
rudigiesler, line 300: I think we also need a test for if there's no clinic for the given code, eg. someone types in the code wrong. In that case Jembi's API will just return an empty array. We handle it here in the jsbox: https://github.com/praekeltfoundation/ndoh-jsbox/blob/develop/src/ussd_clinic.js#L214-L215

registrations/test_views.py
```diff
@@ -235,6 +235,99 @@ def test_get_registration_only_by_user(self):
...
+                urlencode({"criteria": "value:{}".format(clinic_code)})
+            ),
+            json=result,
+            status=200,
+        )
+        url = "{}?{}".format(
+            reverse("facilitycode-check"), urlencode({"clinic_code": clinic_code})
+        )
+        response = self.normalclient.get(url)
+        self.assertEqual(response.status_code, 404)
```
rudigiesler: I'm not sure that we should return a 404 here. If we do, then this goes to the "error" branch of the rapidpro flow, and then it's not easy to know whether the http request failed, or if the user typed in an invalid clinic code.

I think we should instead return a 200, and have the body describe that no clinic was found, either by having a field that's true/false, or by just returning a special value for the name if a clinic isn't found.

OMosimege: could I return something like `resp = {"result": {"Facility": "incorrect clinic code"}`?

rudigiesler: Yup, that would be the "returning a special value" solution. We should probably just make it something like "invalid", I'm not sure how difficult it is to match on a thing with spaces in RapidPro.

registrations/views.py
```diff
@@ -1095,6 +1095,57 @@ def prune(self, request):
...
+                urljoin(settings.JEMBI_BASE_URL, "facilityCheck"),
+                headers={"Content-Type": "application/json"},
+                auth=(settings.JEMBI_USERNAME, settings.JEMBI_PASSWORD),
+                params={"criteria": "value:{}".format(clinic_code)},
+                verify=False,
+            )
+            result.raise_for_status()
+            jembi_result = result.json()
+            if len(jembi_result.get("rows")) > 0:
+                resp = {"result": {"Facility": jembi_result.get("rows")[0][2]}}
```
rudigiesler: We probably want to simplify this a bit, something like `{"facility": "test facility name"}`, I think getting nested fields might be a bit difficult in RapidPro.

OMosimege: okay thanks

registrations/test_views.py
OMosimege: will do

rudigiesler (approved):

PR #223, opened 2019-06-06
Add API for creating clinic registrations from RapidPro

rudigiesler:
This PR adds the serializer, view, and celery tasks required to create a clinic registration from RapidPro


OMosimege (approved):
Looks good 🍰

PR #224, opened 2019-06-07
Fix transmission of context between tasks

rudigiesler:
Currently, the update identity task is not returning the context, which makes the next task, which expects the context, break.

We should fix this, but also create an end-to-end test to ensure that all the tasks are passing along data in a way that the next task understands.

### DIFF
--- a/registrations/test_tasks.py
+++ b/registrations/test_tasks.py
@@ -19,6 +19,7 @@ from registrations.serializers import RegistrationSerializer
 from registrations.signals import psh_validate_subscribe
 from registrations.tasks import (
     _create_rapidpro_clinic_registration,
+    create_rapidpro_clinic_registration,
     get_or_create_identity_from_msisdn,
     get_whatsapp_contact,
     update_identity_from_rapidpro_clinic_registration,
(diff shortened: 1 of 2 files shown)


registrations/test_tasks.py
```diff
@@ -19,6 +19,7 @@
 from registrations.signals import psh_validate_subscribe
 from registrations.tasks import (
     _create_rapidpro_clinic_registration,
+    create_rapidpro_clinic_registration,
```
OMosimege, line 22: what's the difference between this and the one above?

OMosimege (commented):
Looks great 🍰  just one comment

registrations/test_tasks.py
rudigiesler: the `_` is a private task that shouldn't be called directly, that's the task that creates the registration. The other task is the public one that should be called directly, that task fetches the identities, updates the identities, and then runs the `_` task.

OMosimege: ahh, I see

OMosimege (approved):

PR #225, opened 2019-06-10
Add postbirth registrations for MomConnect

rudigiesler:
This modifies the validate_subscribe task to be able to process registrations that happen after the birth of the child.


OMosimege (approved):
great work 👍

PR #226, opened 2019-06-12
Add view for creating public registrations from RapidPro webhooks

rudigiesler:
This is similar to the view for clinic registrations, but is instead focussed on public registrations


OMosimege (approved):
🍰

PR #227, opened 2019-06-14
Add subscription description

rudigiesler:
This PR adds a human readable string description of the current subscriptions that a mother has.

It calculates this based on the history of registrations and changes for the mother.


OMosimege (approved):
Looks great 👍

PR #228, opened 2019-06-14
Split out RapidPro Clinic registration serializer

rudigiesler:
So it turns out to be quite difficult to not have a certain query param in the RapidPro webhook if the user didn't take that path, eg. if they didn't take the passport path, and you do a `passport_origin=@flow.passport_origin.category` in the request, then it just puts the raw text `@flow.passport_origin.category` as the value. This makes the request invalid, as it's not one of the passport choices.

By splitting it out like this, it means that when we submit it, if we have fields that we're submitting that aren't part of the registration or id type, then they'll get ignored instead of being validated.

So back to the example, if the ID type is `sa_id`, then if we have a passport origin specified with some weird text, it will just get ignored, instead of the request failing because of a field that we're not using.


OMosimege (approved):
Looks good 🍰

PR #230, opened 2019-06-18
Feature/momza 1158 filter whatsapp unsent events

githubmonkey:


changes/test_views.py
```diff
@@ -114,6 +114,36 @@ def test_serializer_succeeded(self, task, mock_validate_signature):
             "41c377a47b064eba9abee5a1ea827b3d", user.pk, errors
         )
 
+    def test_filter_failed(self, task, mock_validate_signature):
+        """
+        If the serializer passes but the message does not pass the current filter,
+        then the task should be called with the
```
rudigiesler, line 120: I think this docstring is incomplete

rudigiesler (commented):
Looks great! Just a tiny nitpick

rudigiesler:
This PR is related to the older Seed system, which is no longer in use. It has since been moved over to RapidPro.

PR #231, opened 2019-06-20
Fix RapidPro subscription lookup

rudigiesler:
Currently when there are no changes, we're using `datetime.min` as the limiting timestamp. But if there is an optout, but no baby switch, that will mean that we're comparing a timezone aware timestamp with a timezone unaware one.

We should add a timezone to `datetime.min` to ensure that we're always comparing with timezones.


OMosimege (approved):
Looks good 🍰

PR #232, opened 2019-06-27
Add whatsapp source to authority check

KaitCrawford:
Registrations Via Whatsapp are not being sent through to Jembi.
This PR adds the WhatsApp registration sources to the authority check so that the task doesn't fail when we try to send it to Jembi.


OMosimege (approved):
Looks great 🍰

PR #234, opened 2019-07-01
Use correct mha for momconnect registrations via Whatsapp

KaitCrawford:
We need to send mha=6 to Jembi for registrations via Whatsapp. This value is pulled from the registration data, so when we create a registration from whatsapp we need to add that value to the reg data.


OMosimege (approved):
Looks good 👍

PR #235, opened 2019-07-02
Feature/momza 1236 add event id for jembi

OMosimege:
In this PR, we:

Submit the event ID with every submission to Jembi’s API

Implement automatic retries for when the request to Jembi’s API fails.


changes/tasks.py
```diff
@@ -892,6 +892,10 @@ def check_switch_channel(self, data_fields, change):
             return ["'channel' must be one of {}".format(sorted(channel_types))]
         return []
 
+    def get_change_id(self, change):
```
erikh360: Doesn't look like this gets used, you can remove.

changes/tests.py
```diff
@@ -1875,22 +1878,24 @@ def test_baby_switch_momconnect_only_sub(self):
         utils_tests.mock_patch_identity("mother01-63e2-4acc-9b94-26663b9bc267")
 
         # Execute
+        validate_implement(change.id)
```
erikh360: I don't think we should call this function twice. Thats why you had to double the values on line 1885, 1886:
```
self.assertEqual(SubscriptionRequest.objects.all().count(), 1)
self.assertEqual(len(responses.calls), 9)
```

registrations/tasks.py
```diff
@@ -1025,28 +1025,7 @@ def run(self, registration_id, **kwargs):
...
-                raise self.retry(exc=e)
-            else:
-                self.log.error("Error when posting to Jembi. Payload: %r" % (json_doc))
-                raise e
-        except (Exception,) as e:
-            self.log.error(
-                "Problem posting Registration %s JSON to Jembi" % (registration_id),
-                exc_info=True,
-            )
+        request_to_jembi_api(self.URL, json_doc)
```
erikh360, line 1028: I would change this to: `request_to_jembi_api.delay(self.URL, json_doc)`, the delay part makes sure it is not executed directly but put on the queue.

registrations/tests.py
```diff
@@ -2810,7 +2810,7 @@ def test_registration_process_pmtct_good(self):
         # . check number of calls made:
         #   messageset, schedule, identity, patch identity, jembi registration
         #   identity, reverse identity, reverse identity, patch identity
-        self.assertEqual(len(responses.calls), 9)
```
erikh360: This is a bit weird, we didn't remove any calls did we?

Let investigate by looking at each call and updating the comment above this line.

registrations/tests.py
```diff
@@ -2927,7 +2928,7 @@ def test_registration_process_pmtct_no_facility(self):
         Registration.objects.create(**registration_data)
 
         # check jembi registration
-        jembi_call = responses.calls[13]  # jembi should be the fourteenth one
+        jembi_call = responses.calls[9]  # jembi should be the tenth one
```
erikh360: All these where the calls are less doesn't look right.

erikh360 (commented):
A few comments, but looks good overall.

changes/tasks.py
OMosimege: thanks

registrations/tasks.py
OMosimege: because in the run function that executes this, the task is done by task.delay(), wont this result in a double delay?

registrations/tests.py
OMosimege: we did remove the try and except, which looked to have (3 default retries)

registrations/tasks.py
erikh360: No, it just means the main task can put this one on the queue and complete. Then the celery worker will pick up the `request_to_jembi_api` task and execute.

registrations/tests.py
OMosimege: may my above comment relate to this as well?

registrations/tests.py
erikh360: The whole try and except block was only doing one post, which was mocked so it should never have executed the except parts.

changes/tests.py
```diff
@@ -1755,12 +1755,14 @@ def test_baby_switch_momconnect_pmtct_nurseconnect_subs(self):
         self.assertEqual(Registration.objects.all().count(), 0)
         # Setup
         # make registrations
+        url = "http://jembi/ws/rest/v1/subscription"
```
erikh360: Can we rename this to `jembi_url`?

registrations/tests.py
erikh360: Can we double check this one please?

I recon it's because it wasn't mocked so it retried 5 times.

registrations/tests.py
```diff
@@ -3072,10 +3074,8 @@ def test_registration_process_clinic_good(self):
 
         # Check
         # . check number of calls made:
-        #   message set, schedule, popi message set, jembi registration,
-        #   id_store mother, id_store mother_reverse,
-        #   id_store registrant_rever, id_store patch
-        self.assertEqual(len(responses.calls), 8)
+        #   message set, schedule, popi message set, jembi registration
+        self.assertEqual(len(responses.calls), 4)
```
erikh360: Is this number ok? Looks like we're not doing the requests from `remove_personally_identifiable_fields` anymore?

erikh360 (commented):
Just a few comments and I replied on comments from the last review.

The url -> jembi_url change is not very important but good for readability. I see it's in a few places.

changes/tests.py
OMosimege: done

registrations/tests.py
OMosimege: it was missing the request_jembi_url mock

erikh360 (approved):
Awesome!! 🎉

PR #236, opened 2019-07-02
Feature/momza 1237 add system id for jembi

OMosimege:
In this PR we submit a system id `sid` with new registrations and events to Jembi’s API, as the identity/contact’s UUID in the information that we submit to them


rudigiesler (approved):
👍 Looks good

I think we just need to make sure that we're making this change wherever we've added the `eid` field. I think we're missing some places, eg. sending the helpdesk message.

PR #237, opened 2019-07-05
added request_jembi_api task to views

OMosimege:


registrations/tasks.py
```diff
@@ -1543,3 +1543,4 @@ def request_to_jembi_api(url, json_doc):
         verify=False,
     )
     r.raise_for_status()
```
erikh360: You can probably remove this, we don't want to `raise_for_status` twice.

registrations/views.py
```diff
@@ -314,6 +315,7 @@ def get_software_type(channel_id):
             or self.UNCLASSIFIED_MESSAGES_DEFAULT_LABEL,
             "type": 7,  # 7 helpdesk
             "op": str(validated_data.get("helpdesk_operator_id")),
+            "eid": str(registration.id) if registration else None,
```
erikh360: I don't think we should use the registration ID here.

If the idea is to use this field to deduplicate events sent to jembi, multiple incoming messages from this registration will be the same, therefore ignored.

We should rather be using something unique to the incoming message, like a message ID.

registrations/views.py
```diff
@@ -420,7 +398,6 @@ def get(self, request, *args, **kwargs):
 
         except (requests.exceptions.HTTPError,) as e:
             if e.response.status_code == 400:
-                print(e.response.text)
```
erikh360, line 423: You can remove this 🙂

registrations/views.py
```diff
@@ -332,39 +334,15 @@ def post(self, request):
...
-                )
-            else:
-                raise e
+        source = Source.objects.get(user=self.request.user.id)
+        endpoint = "helpdesk"
+        if source.name == "NURSE Helpdesk App":
+            endpoint = "nc/helpdesk"
+            post_data["type"] = 12  # NC Helpdesk
+        jembi_url = urljoin(settings.JEMBI_BASE_URL, endpoint)
+        raise_status = request_to_jembi_api(jembi_url, post_data)
```
erikh360: We probably don't need to return the `raise_for_status` here. Any error would cause an exception triggering our new retry strategy.

On success we can just return the 200 as it was, because the only way it will get to the return line is if nothing went wrong.

erikh360 (dismissed):
Requested a few changes 🙂

registrations/tasks.py
OMosimege: okay thanks

registrations/views.py
OMosimege: ahh I missed this, thanks

OMosimege: oh wait, removed it already

changes/tasks.py
```diff
@@ -1860,6 +1860,7 @@ def get_engage_inbound_and_reply(wa_contact_id, message_id):
     inbound_text = " | ".join(list(inbound_text)[::-1])
     labels = map(lambda m: m["_vnd"]["v1"]["labels"], inbounds)
     labels = map(lambda l: l["value"], ichain.from_iterable(labels))
+    message_id = message_id
```
rudigiesler: This doesn't really do anything, so we can probably remove this line.

registrations/tests.py
```diff
@@ -3945,10 +3958,14 @@ def test_send_outgoing_message_to_jembi_with_null_operator_id(self):
 
     @responses.activate
     def test_send_outgoing_message_to_jembi_invalid_user_id(self):
-        self.make_source_normaluser()
+        message_id = "BCGGJ3FVFUV"
+        jembi_url = "http://jembi/ws/rest/v1/helpdesk"
+        source = self.make_source_normaluser()
+        self.make_registration_for_jembi_helpdesk(source)
+        utils_tests.mock_request_to_jembi_api(jembi_url)
```
rudigiesler: I don't think we need all these extra bits in here (specifically the registration and the additional `mock_request_to_jembi_api`)

registrations/views.py
```diff
@@ -332,37 +334,13 @@ def post(self, request):
...
-                )
-            else:
-                raise e
+        source = Source.objects.get(user=self.request.user.id)
+        endpoint = "helpdesk"
+        if source.name == "NURSE Helpdesk App":
+            endpoint = "nc/helpdesk"
+            post_data["type"] = 12  # NC Helpdesk
+        jembi_url = urljoin(settings.JEMBI_BASE_URL, endpoint)
+        request_to_jembi_api(jembi_url, post_data)
```
rudigiesler: This isn't something specifically for this PR, but since we're modifying this code anyway. I've always wanted this HTTP call to be done asynchronously, that way we can do retries and better error handling, and we don't block the HTTP worker from processing requests if Jembi's API is slow or down.

It should be as simple as adding a `.delay`, ie. `request_to_jembi_api.delay(jembi_url, post_data)`

This will probably mean that the one test will start failing, because we'll no longer be returning an error response if Jembi's API gives us an error (we can't since it's no longer synchronous). I'm happy to just remove that test though.

registrations/serializers.py
```diff
@@ -313,6 +313,7 @@ class JembiHelpdeskOutgoingSerializer(serializers.Serializer):
     inbound_created_on = serializers.DateTimeField()
     outbound_created_on = serializers.DateTimeField()
     inbound_channel_id = serializers.CharField(required=False, allow_blank=True)
+    message_id = serializers.CharField()
```
rudigiesler, line 316: Nothing to change here, but we just need to remember to update CasePro to send this extra field.

rudigiesler (changes requested):

registrations/serializers.py
OMosimege: so I need to remove this?

registrations/tests.py
OMosimege: okay will do

registrations/views.py
OMosimege: if it affects 2 tests that I had to remove, is this still okay?

changes/tasks.py
OMosimege: noted

registrations/serializers.py
rudigiesler: No, nothing needs to change in this PR. We need to change CasePro to send this field to us, because currently CasePro isn't sending us the message ID.

registrations/views.py
rudigiesler: Yes that should be fine. I'll check that it's okay once you push the two removed tests.

registrations/serializers.py
OMosimege: okay noted

rudigiesler (approved):

PR #238, opened 2019-07-16
Optin after registration

rudigiesler:
Currently, if someone opts out, and then registers again, they don't get opted in again, so they don't receive any messages.

We should check if their number is opted out, and if so, opt them in again after a successful registration, to ensure that they can receive messages for their new registration.


OMosimege (approved):
Looks good

PR #239, opened 2019-07-16
Send welcome message after successful prebirth registration

rudigiesler:
Currently the USSD line manually sends the welcome message, but this means that other registration methods (specifically registration via WhatsApp) do not get a welcome SMS.

This PR ensures that all successful registrations get a welcome SMS.

There will also be a PR in https://github.com/praekeltfoundation/ndoh-jsbox/ to stop the USSD line from sending a welcome message.


OMosimege (approved):
looks good 👍

PR #242, opened 2019-07-25
converted messade id to string uuid from integer

OMosimege:


changes/test_tasks.py
```diff
@@ -521,7 +522,7 @@ def test_no_identity_found(self):
 class GetEngageInboundAndReplyTests(TestCase):
     @responses.activate
     def test_get_engage_inbound_and_reply(self):
-        message_id = "BCGGJ3FVFUV"
+        message_id = 10
```
rudigiesler: These message IDs that we're getting from Turn aren't integers, so we can't make them integers in this test. We get back IDs that look like `ABGGJ3NndzVPAhAoVGKeRSJmvqOOjqbFI-wC`, which I'm not sure that we can really use. 

I think the next best thing would be to use the celery task ID. This won't protect us against whatsapp sending us the message twice, but we weren't protected against that beforehand anyway.

If you set `bind=True` in the `@app.task(...)`, then the first argument passed to the task will be `self`. You can then use `self.request.id` to get the ID of the task, which is a UUID

rudigiesler (changes requested):
This PR seems to tackle the issue of when we need to send the event from the WhatsApp helpdesk (turn), but it doesn't seem to tackle the issue where we get an event from the SMS helpdesk (casepro). That is handled in https://github.com/praekeltfoundation/ndoh-hub/blob/develop/registrations/views.py#L302-L320

changes/test_tasks.py
```diff
@@ -521,7 +521,7 @@ def test_no_identity_found(self):
 class GetEngageInboundAndReplyTests(TestCase):
     @responses.activate
     def test_get_engage_inbound_and_reply(self):
-        message_id = "BCGGJ3FVFUV"
+        message_id = "cdffd588-dc29-469d-b2ac-3a0c2d5d8609"
```
rudigiesler: I don't think we should be changing these message IDs to UUIDs, since in practice they are not, so anything that assumes that they are UUIDs we want that to break in the tests.

changes/test_tasks.py
```diff
@@ -1030,7 +1032,9 @@ def assert_openhim_request(request):
             registrant_id="identity-uuid", data={"faccode": "123456"}, source=source
         )
 
-        process_engage_helpdesk_outbound.delay("27820001001", message_id).get()
+        process_engage_helpdesk_outbound.apply_async(
+            args=["27820001001", message_id], task_id=message_id
```
rudigiesler, line 1037: This shouldn't be the message ID, this should be a UUID. We want the message ID and the task UUID to be two different things.

rudigiesler (approved):

PR #245, opened 2019-08-16
Handle null message type when processing whatsapp messages

rudigiesler:
For older messages, they don't have a message type, so we need to be able to handle that.


OMosimege (approved):
Looks good 👍

PR #246, opened 2019-08-22
Add script for getting archived messages

rudigiesler:
This script will allow us to pull the archived messages, and filter them to only get the ones that we're interested in


OMosimege (approved):
good work 👍

PR #248, opened 2019-09-09
Add baby message switch to the event store

rudigiesler:
This allows us to store switches to baby messaging events in the event store


OMosimege (approved):
great work👍

PR #250, opened 2019-09-20
Add public registrations to the event store

rudigiesler:
This allows us to store the details of all public registrations in the event store


OMosimege (approved):
Looks good 👍

PR #253, opened 2019-10-03
Denormalise created_by on eventstore

rudigiesler:
For this eventstore, we probably don't want any relations, since that adds to the storage and performance costs.

We also will never use the relation, it's just for historical purposes, not to actually query the events created by a specific user.


OMosimege (approved):
Everything looks great 👍 Just wanted to ask why we are denormalising the `created_by`? with history of why it was normalised in the beginning

rudigiesler:
@OMosimege I tried to explain this in the description.

So usually what you want to do with a relational database is that you want to normalise it, using foreign keys, so that you don't end up with inconsistencies in your database, eg. if we had to change the name of a user, then all the old entries will have the old name, and all the new entries would have the new name, so now you have a bunch of old entries created by a user that doesn't exist. So I went ahead and created it like this, because it's usually good practice and what you want to do.

The issue with this is, that for each relation, it creates a database index, so that when you do things like "give me all the events created by user x", it's performant. In this use case, however, we won't ever want to do queries of that kind. These indexes take up disk space, and add a cost for everytime you want to add a row, so if we don't need it, we probably shouldn't have it. Also, since this is a historical event store, if a username changes, we want it to show the old username for old entries, we want to maintain that history.

So thinking about this a bit more, for this specific case, we don't want our data to be normalised, and it was probably a mistake to have it normalised in the first place.

PR #255, opened 2019-10-22
Third party registrations sending to RapidPro

rudigiesler:
This PR sends third party registrations to RapidPro if the correct settings are set


OMosimege (approved):
great work 🍰

PR #256, opened 2019-10-24
Fix msisdn lookups and tests for turn context

rudigiesler:
The responses mocks in the unit tests aren't using match_querystring, which is causing them to pass when there's a bug in the code. This PR fixes the responses mocks, and the bug in the code


OMosimege (commented):
👍

PR #257, opened 2019-10-24
Fix submission to RP on third party registrations

rudigiesler:
The `urns` should be a list not a string, and we should also do the task in the background, not in-process


OMosimege (approved):
🍰

PR #258, opened 2019-10-25
Add CHW registrations to the event store

rudigiesler:
This PR allows us to store CHW registrations in the event store


OMosimege (approved):
👍

PR #260, opened 2019-10-30
Add API for nurseconnect facility check

rudigiesler:


KaitCrawford (approved):
:+1: looks good to me
Are we sure the NC and MC facility checks have the same set of clinics though?

rudigiesler:
@KaitCrawford They have a slightly different set of codes, but for the purpose of this feature I'm just going to merge it into one, I'm not sure why they're different
```
➜  ~ comm -1 <(sort momcodes.csv) <(sort nursecodes.csv) | wc -l
   19654
➜  ~ comm -13 <(sort momcodes.csv) <(sort nursecodes.csv) | wc -l
      64
```

PR #261, opened 2019-10-31
Store jembi requests before forwarding

rudigiesler:
The PR adds the functionality to store the events that will be sent to jembi's API before sending them. This is part of the functionality that will allow us to turn off the forwarding and forward them at a later stage


OMosimege (approved):
👍

PR #264, opened 2019-11-01
Make sending events to Jembi optional

rudigiesler:
This has 2 parts.

First, we need to have a flag that we can set to stop sending events.

Second, we need a command we can run to resend any events that weren't sent


registrations/tests.py
```diff
@@ -3869,6 +3723,36 @@ def test_send_outgoing_message_to_jembi(self):
...
+            "inbound_channel_id": "6a5c691e-140c-48b0-9f39-a53d4951d7fa",
+            "message_id": message_id,
+        }
+        # Execute
+        response = self.normalclient.post(
+            "/api/v1/jembi/helpdesk/outgoing/", user_request
+        )
+
+        self.assertEqual(response.status_code, 200)
+        self.assertEqual(len(responses.calls), 1)
```
KaitCrawford, line 3754: What is the call being made here?

rudigiesler: It's the call to Junebug, so probably not the best test to pick, but you can see the previous test asserts 2 calls, and this one asserts one

KaitCrawford: ok

KaitCrawford (approved):
:+1:

PR #265, opened 2019-11-04
Feature/mcmr 80 store messages and events in event store

OMosimege:
This PR creates a model for Events and Messages, and creates a view to handle them


eventstore/models.py
```diff
@@ -160,3 +160,27 @@ class PostbirthRegistration(models.Model):
     timestamp = models.DateTimeField(default=timezone.now)
     created_by = models.CharField(max_length=255, blank=True, default="")
     data = JSONField(default=dict, blank=True, null=True)
+
+
+class Messages(models.Model):
```
rudigiesler: It's considered Django best practice to only name models using the singular, not plural: https://wsvincent.com/django-models-best-practices/

eventstore/models.py
```diff
@@ -160,3 +160,27 @@ class PostbirthRegistration(models.Model):
...
+    INBOUND = "I"
+    OUTBOUND = "O"
+    DIRECTION_TYPES = [(INBOUND, "Inbound"), (OUTBOUND, "Outbound")]
+
+    id = models.CharField(max_length=255, primary_key=True, blank=True)
+    contact_id = models.CharField(max_length=255, blank=True)
+    timestamp = models.DateTimeField(default=timezone.now)
+    type = models.CharField(max_length=255, blank=True)
+    data = JSONField(default=dict, blank=True, null=True)
+    message_direction = models.CharField(max_length=3, choices=DIRECTION_TYPES)
```
rudigiesler: None of our types have a length longer than 1, so we can reduce the `max_length` here to 1

eventstore/test_views.py
```diff
@@ -387,6 +394,161 @@ def test_postbirth_other_passport_origin(self):
...
+        user.user_permissions.add(Permission.objects.get(codename="add_messages"))
+        self.client.force_authenticate(user)
+        data = {
+            "messages": [
+                {
+                    "id": "9e12d04c-af25-40b6-aa4f-57c72e8e3f91",
+                    "from": "sender-wa-id",
+                    "timestamp": "1518694700",
+                    "type": "image",
+                    "message_direction": "I",
```
rudigiesler: The `message_direction` is not a field that WhatsApp will be sending us in this webhook, so we should remove it here

eventstore/test_views.py
```diff
@@ -387,6 +394,161 @@ def test_postbirth_other_passport_origin(self):
...
+                    "recipient_type": "type",
```
rudigiesler: The `recipient_type` is not a field that WhatsApp would send us, so we should remove it

eventstore/test_views.py
```diff
@@ -387,6 +394,161 @@ def test_postbirth_other_passport_origin(self):
...
+                    "system": {"body": "system-message-content"},
+                    "text": {"body": "text-message-content"},
+                }
+            ]
+        }
+        response = self.client.post(
+            self.url,
+            data,
+            format="json",
+            HTTP_X_TURN_HOOK_SIGNATURE=self.generate_hmac_signature(data, "REPLACEME"),
```
rudigiesler, line 647: There's also the `X-Turn-Hook-Subscription` header that will be sent to us.

eventstore/test_views.py
```diff
@@ -387,6 +394,161 @@ def test_postbirth_other_passport_origin(self):
...
+
+        self.assertEqual(response.status_code, status.HTTP_202_ACCEPTED)
+        [messages] = Messages.objects.all()
+        self.assertEqual(str(messages.contact_id), "sender-wa-id")
+        self.assertEqual(
+            messages.timestamp, datetime.datetime(2018, 2, 15, 11, 38, 20, tzinfo=UTC)
+        ),
+        self.assertEqual(messages.id, "9e12d04c-af25-40b6-aa4f-57c72e8e3f91"),
+        self.assertEqual(messages.type, "image"),
+        self.assertEqual(messages.message_direction, "I"),
```
rudigiesler: We should probably use `Messages.INBOUND` here, instead of the magic number `"I"`

eventstore/test_views.py
```diff
@@ -387,6 +394,161 @@ def test_postbirth_other_passport_origin(self):
...
+                        "mime_type": "media-mime-type",
+                        "sha256": "checksum",
+                        "caption": "image-caption",
+                    },
+                    "location": {
+                        "address": "1 Hacker Way, Menlo Park, CA, 94025",
+                        "name": "location-name",
+                    },
+                    "system": {"body": "system-message-content"},
+                    "text": {"body": "text-message-content"},
```
rudigiesler, line 637: The webhook API format changes often, so I think we should harden our API against this.

In this test, we could add some `"random": "data",`, and ensure that it gets placed inside the `data` field on our model, to simulate WhatsApp adding a field to the request that we're not expecting.

eventstore/test_views.py
```diff
@@ -387,6 +394,161 @@ def test_postbirth_other_passport_origin(self):
...
+                "location": {
+                    "address": "1 Hacker Way, Menlo Park, CA, 94025",
+                    "name": "location-name",
+                },
+                "system": {"body": "system-message-content"},
+                "text": {"body": "text-message-content"},
+            },
+        ),
+        self.assertEqual(messages.created_by, user.username)
+        self.assertEqual(messages.recipient_type, "type")
```
rudigiesler: I'm not sure what the `recipient_type` field is, but there's no mention of it in the WhatsApp documentation. 

Maybe we can remove this field from the model?

eventstore/test_views.py
```diff
@@ -387,6 +394,161 @@ def test_postbirth_other_passport_origin(self):
...
+        user = get_user_model().objects.create_user("test")
+        user.user_permissions.add(Permission.objects.get(codename="add_messages"))
+        self.client.force_authenticate(user)
+        data = {
+            "statuses": [
+                {
+                    "id": "ABGGFlA5FpafAgo6tHcNmNjXmuSf",
+                    "recipient_id": "16315555555",
+                    "status": "read",
+                    "timestamp": "1518694700",
```
rudigiesler, line 752: Same here with adding some `"random": "data",`, and ensuring that it gets put inside the `data` of the model

eventstore/test_views.py
```diff
@@ -387,6 +394,161 @@ def test_postbirth_other_passport_origin(self):
...
+                }
+            ]
+        }
+        response = self.client.post(
+            self.url,
+            data,
+            format="json",
+            HTTP_X_TURN_HOOK_SIGNATURE=self.generate_hmac_signature(data, "REPLACEME"),
```
rudigiesler, line 761: Here as well for the `X-Turn-Hook-Subscription` header that we receive

eventstore/test_views.py
```diff
@@ -387,6 +394,161 @@ def test_postbirth_other_passport_origin(self):
...
+                    "message_direction": "I",
+                    "recipient_type": "type",
+                }
+            ]
+        }
+        response = self.client.post(self.url, data, format="json")
+
+        self.assertEqual(response.status_code, status.HTTP_401_UNAUTHORIZED)
+
+
```
rudigiesler, line 807: It seems like we're missing a test for outbound messages

rudigiesler, line 807: We should also add a test that if the user doesn't have the required permission, then they're not allowed to access the endpoint

rudigiesler, line 807: We're using the `force_authentication` method for our tests, which makes testing easy, but we should add a test that doesn't have a token in the querystring, and test that we get an auth errors, and then puts the token of the user in the querystring, and tests that the request works fine, because that's how this API endpoint is going to be used.

eventstore/views.py
```diff
@@ -20,6 +27,89 @@
...
+                Messages.objects.create(
+                    id=request.headers["X-WhatsApp-Id"],
+                    contact_id=inbound.to,
+                    type=type,
+                    data=data,
+                    message_direction=inbound.OUTBOUND,
+                    created_by=request.user.username,
+                    recipient_type=inbound.recipient_type,
+                )
+        except ValidationError:
```
rudigiesler: I'm not seeing anything in this code block that would trigger a ValidationError, and we don't have any tests that would be calling this part of the code, so we can probably remove this `try..catch` block

eventstore/views.py
```diff
@@ -20,6 +27,89 @@
...
     PublicRegistrationSerializer,
 )
+from ndoh_hub.utils import validate_signature
+
+
+class MessagesViewSet(GenericViewSet):
+    def create(self, request):
+        validate_signature(request)
+        try:
+            webhook_type = request.META.get("HTTP_X_TURN_HOOK_SUBSCRIPTION", "whatsapp")
```
rudigiesler: Here, we're defaulting to a `whatsapp` webhook type if the header isn't included.

I don't think that this is the correct way to do this, we should rather send a 400 Bad Request error back if the header isn't specified. (and also add a unit test that tests this)

eventstore/views.py
```diff
@@ -20,6 +27,89 @@
...
+                        "errors": inbound.get("errors"),
+                        "audio": inbound.get("audio"),
+                        "document": inbound.get("document"),
+                        "image": inbound.get("image"),
+                        "system": inbound.get("system"),
+                        "location": inbound.get("location"),
+                        "text": inbound.get("text"),
+                        "video": inbound.get("video"),
+                        "voice": inbound.get("voice"),
+                    }
```
rudigiesler: Doing it this way won't allow us to easily adapt to new fields being added to the message body, we would have to change this block of code to be able to adapt to changes in the webhook.

Perhaps a better way of doing this is to use Python's `pop` method on the dictionary, ie. For each of the required fields, we do a `pop`
```
id = inbound.pop("id")
contact_id = inbound.pop("from")
...
```
Then, anything remaining in `inbound`, we can put into `data` on the `Messages`, ie.
```
Messages.objects.create(
  id=id,
  contact_id=contact_id,
  ...,
  data=inbound
)
```

eventstore/views.py
```diff
@@ -20,6 +27,89 @@
...
+                        type=inbound["type"],
+                        data=data,
+                        message_direction=Messages.INBOUND,
+                        created_by=request.user.username,
+                        recipient_type=inbound["recipient_type"],
+                        timestamp=datetime.fromtimestamp(int(inbound["timestamp"])),
+                    )
+
+                for statuses in request.data.get("statuses", []):
+                    data = {"errors": statuses.get("errors")}
```
rudigiesler: Same here for rather using `pop` to get the required fields, and then putting the remaining data in `data`

eventstore/views.py
```diff
@@ -20,6 +27,89 @@
...
+                        "document": inbound.get("document"),
+                        "image": inbound.get("image"),
+                        "system": inbound.get("system"),
+                        "location": inbound.get("location"),
+                        "text": inbound.get("text"),
+                        "video": inbound.get("video"),
+                        "voice": inbound.get("voice"),
+                    }
+                    Messages.objects.create(
+                        id=inbound["id"],
```
rudigiesler: This will raise a `KeyError` if the field doesn't exist, which we're not catching, which will mean we'll get an error in Sentry, and response with a 500 status code.

Instead, we should rather catch the key error, and respond with a 400 Bad Request

eventstore/test_views.py
rudigiesler, line 807: We have seen that WhatsApp sends us the same messages multiple times. We should create a test to ensure that, if we're being sent the same message twice, that we can handle that without an error.

eventstore/views.py
```diff
@@ -20,6 +27,89 @@
...
+                }
+                Messages.objects.create(
+                    id=request.headers["X-WhatsApp-Id"],
+                    contact_id=inbound.to,
+                    type=type,
+                    data=data,
+                    message_direction=inbound.OUTBOUND,
+                    created_by=request.user.username,
+                    recipient_type=inbound.recipient_type,
+                )
```
rudigiesler, line 95: This seems to be just a copy of the previous code block.

I think instead of trying to parse the request, if the header doesn't contain a value that we recognise, we should return a 400 Bad Request error, explaining that we only accept `whatsapp` and `turn` webhook types.

ndoh_hub/utils.py
```diff
@@ -65,6 +69,21 @@ def get_identity_msisdn(registrant_id):
...
+def validate_signature(request):
+    secret = settings.TURN_CONTEXT_HMAC_SECRET
+    try:
+        signature = request.META["HTTP_X_TURN_HOOK_SIGNATURE"]
+    except KeyError:
+        raise AuthenticationFailed("X-Turn-Hook-Signature header required")
+
+    h = hmac.new(secret.encode(), request.body, sha256)
+    print(request.body)
+    print(secret.encode())
```
rudigiesler: We can remove these print statements

eventstore/views.py
rudigiesler: We should probably use `request.headers` here as well, so that we're consistent and not using `request.headers` in some places, but `request.META` in others.

ndoh_hub/settings.py
```diff
@@ -279,6 +279,7 @@
 
 ENGAGE_HMAC_SECRET = os.environ.get("ENGAGE_HMAC_SECRET", "REPLACEME")
 ENGAGE_CONTEXT_HMAC_SECRET = os.environ.get("ENGAGE_CONTEXT_HMAC_SECRET", "REPLACEME")
+TURN_CONTEXT_HMAC_SECRET = os.environ.get("ENGAGE_CONTEXT_HMAC_SECRET", "REPLACEME")
```
rudigiesler: I think we should rather have this be `TURN_HMAC_SECRET` and get it from the `TURN_HMAC_SECRET` environment variable, since we're not using it for the context API here.

eventstore/models.py
OMosimege: Okay, thanks for the resource

eventstore/models.py
OMosimege: okay will do

eventstore/test_views.py
OMosimege: noted

eventstore/test_views.py
OMosimege: okay noted

ndoh_hub/settings.py
OMosimege: okay will do

ndoh_hub/utils.py
OMosimege: ahh forgot those, thanks

eventstore/views.py
OMosimege: so, under the else, I could simple return a `return Response(status=status.HTTP_400_BAD_ERROR)` ?

rudigiesler: Yes, we'll probably also want to include a body explaining why we sent a 400

eventstore/test_views.py
OMosimege: okay, will create an outbound test as well

eventstore/test_views.py
OMosimege: okay, removed

eventstore/test_views.py
OMosimege: done

eventstore/test_views.py
OMosimege: created it, thanks

eventstore/views.py
OMosimege: will try this out

eventstore/test_views.py
OMosimege: okay, thanks

eventstore/test_views.py
OMosimege: will add it

eventstore/views.py
OMosimege: done

eventstore/test_views.py
OMosimege: done, thanks for the assistance

eventstore/test_views.py
```diff
@@ -387,6 +395,284 @@ def test_postbirth_other_passport_origin(self):
...
+        user = get_user_model().objects.create_user("test")
+        user.user_permissions.add(Permission.objects.get(codename="add_message"))
+        data = {"random": "data"}
+
+        response = self.client.post(
+            self.url,
+            data,
+            format="json",
+            HTTP_X_TURN_HOOK_SIGNATURE=self.generate_hmac_signature(data, "REPLACEME"),
+        )
```
rudigiesler, line 452: We're not asserting anything on this request. We can either remove it, or assert something useful about it (eg. that we get an auth error)

eventstore/test_views.py
```diff
@@ -387,6 +395,284 @@ def test_postbirth_other_passport_origin(self):
...
+        }
+        response = self.client.post(
+            self.url,
+            data,
+            format="json",
+            HTTP_X_TURN_HOOK_SIGNATURE=self.generate_hmac_signature(data, "REPLACEME"),
+            HTTP_X_TURN_HOOK_SUBSCRIPTION="other",
+        )
+
+        self.assertEqual(response.status_code, status.HTTP_400_BAD_REQUEST)
```
rudigiesler, line 495: We're not checking that we're explaining why we're returning the error in the response body, we should probably do that

eventstore/views.py
```diff
@@ -20,6 +27,67 @@
...
+
+    def create(self, request):
+        validate_signature(request)
+        webhook_type = request.headers.get("X-Turn-Hook-Subscription", None)
+        if webhook_type == "whatsapp":
+            for inbound in request.data.get("messages", []):
+                id = inbound.pop("id")
+                contact_id = inbound.pop("from")
+                type = inbound.pop("type")
+                timestamp = datetime.fromtimestamp(int(inbound.pop("timestamp")))
```
rudigiesler, line 56: If any of these fields are missing, that means that this part of the code will break. We probably want to catch that error and return a 400 status back, letting the client know that they didn't supply one of the required fields.

eventstore/views.py
```diff
@@ -20,6 +27,67 @@
...
+                    message_direction=Message.INBOUND,
+                    created_by=request.user.username,
+                    timestamp=timestamp,
+                )
+
+            for statuses in request.data.get("statuses", []):
+                message_id = statuses.pop("id")
+                recipient_id = statuses.pop("recipient_id")
+                timestamp = datetime.fromtimestamp(int(statuses.pop("timestamp")))
+                message_status = statuses.pop("status")
```
rudigiesler, line 74: Same comment here around missing fields

eventstore/views.py
```diff
@@ -20,6 +27,67 @@
...
+                "preview_url": outbound["preview_url"],
+            }
+            Message.objects.create(
+                id=request.headers["X-WhatsApp-Id"],
+                contact_id=outbound["to"],
+                type=outbound["type"],
+                data=data,
+                message_direction=Message.OUTBOUND,
+                created_by=request.user.username,
+                recipient_type=outbound["recipient_type"],
```
rudigiesler: `recipient_type` is not a required field, the way this is currently is that this view will break if `recipient_type` isn't supplied.

eventstore/views.py
```diff
@@ -20,6 +27,67 @@
...
+                    data=statuses,
+                )
+
+        elif webhook_type == "turn":
+            outbound = request.data
+            data = {
+                "text": outbound["text"],
+                "render_mentions": outbound["render_mentions"],
+                "preview_url": outbound["preview_url"],
+            }
```
rudigiesler: This way doesn't allow us to catch random additional fields. We will also need to use the `.pop` method here.

We can test this by adding a `"random": "data"` in the test, and checking that it appears in `data` in the test.

eventstore/views.py
```diff
@@ -20,6 +27,67 @@
...
+                id=request.headers["X-WhatsApp-Id"],
+                contact_id=outbound["to"],
+                type=outbound["type"],
+                data=data,
+                message_direction=Message.OUTBOUND,
+                created_by=request.user.username,
+                recipient_type=outbound["recipient_type"],
+            )
+        else:
+            return Response(status=status.HTTP_400_BAD_REQUEST)
```
rudigiesler: We should probably explain here to the user what they did wrong

eventstore/test_views.py
OMosimege: forgot to add the line, thanks for pointing it out

eventstore/views.py
OMosimege: okay will do

eventstore/views.py
OMosimege: forgot to remove this, will remove

eventstore/views.py
OMosimege: what is the best way to explain to the user what went wrong?

eventstore/views.py
OMosimege: will do

erikh360 (approved):

PR #266, opened 2019-11-19
Send jembi app registrations to RapidPro

rudigiesler:
If the flag is set, then we should instead trigger a flow using the provided information, rather than creating a registration for Seed


erikh360 (approved):
Looks good!

PR #268, opened 2019-11-26
Send external ID to RapidPro for Jembi registration

rudigiesler:
We need this external ID for when we're doing the success/failure callbacks


OMosimege (approved):
👍

PR #270, opened 2019-12-03
Trigger RapidPro flow on receiving operator replies

rudigiesler:
This will allow us to take actions on this event, like submitting the message to OpenHIM


OMosimege (approved):
👍 great work

PR #276, opened 2020-01-13
Remove contact PII on delete optout

rudigiesler:
This will anonymise the data that we store when the contact asks us to delete the data we have on them


OMosimege (approved):
👍

PR #278, opened 2020-01-23
Update Rapidpro preferred channel on WhatsApp inbound

erikh360:
Turn disabled the fallback flag when a user send in a WhatsApp message, this is to keep it in sync in Rapidpro.

### DIFF
--- a/eventstore/whatsapp_actions.py
+++ b/eventstore/whatsapp_actions.py
@@ -2,7 +2,7 @@ from celery import chain
 from django.conf import settings
 
 from changes.tasks import get_engage_inbound_and_reply
-from eventstore.tasks import async_create_flow_start
+from eventstore.tasks import async_create_flow_start, update_rapidpro_contact
 from ndoh_hub.utils import normalise_msisdn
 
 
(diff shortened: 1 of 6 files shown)


eventstore/whatsapp_actions.py
```diff
@@ -26,3 +26,20 @@ def handle_operator_message(message):
...
+def handle_inbound(message):
+    """
+    Triggers all the actions that are required for this inbound message
+    """
+    if not message.fallback_channel:
+        update_rapidpro_preferred_channel(message)
+
+
+def update_rapidpro_preferred_channel(message):
+    whatsapp_contact_id = message.data["_vnd"]["v1"]["chat"]["owner"]
```
rudigiesler: Is Turn always guaranteed to have this key in inbounds? I know we have a bunch of inbound messages in our eventstore without this, but that might've been a bug that has since been fixed.

rudigiesler (approved):

eventstore/whatsapp_actions.py
erikh360: Oh thanks, I'll need to confirm this.

erikh360:
@rudigiesler mind having another look? I'm using the "from" value to construct a whatsapp urn

rudigiesler (approved):

PR #279, opened 2020-01-24
Save inbound if turn event header is set

erikh360:
These turn events comes from the fallback channel so the subscription header wont be `whatsapp`


OMosimege (approved):
👍

PR #280, opened 2020-01-24
Add script for extracting all Seed data that we'll need in RapidPro

rudigiesler:
This will extract all the raw data that we need from the various databases, and place it into a file.


OMosimege (approved):
👍

PR #282, opened 2020-01-28
Handle WhatsApp HSM errors

erikh360:
There is one TODO that I will do in the next PR.

I also updated the isort config and added pre-commit. Sorry for doing this in the same PR


OMosimege (approved):
👍

eventstore/whatsapp_actions.py
```diff
@@ -40,3 +40,32 @@ def update_rapidpro_preferred_channel(message):
...
+                hsm_error = True
+            if "envelope mismatch" in error["title"]:
+                hsm_error = True
+
+        if hsm_error:
+            async_create_flow_start.delay(
+                extra={},
+                flow=settings.RAPIDPRO_UNSENT_EVENT_FLOW,
+                urns=[f"whatsapp:{event.recipient_id}"],
+            )
```
rudigiesler: I'm wondering if this logic should be inside the `Event` model instance, so that we can do an `if event.is_hsm_error: ...` in here.

rudigiesler (commented):

eventstore/whatsapp_actions.py
erikh360: I like this idea, would make it less complex. I'm going to implement

rudigiesler (approved):

PR #284, opened 2020-01-28
Seed migration to RapidPro scripts

rudigiesler:
These scripts will allow us to extract all the required information from Seed, and then upload that information to RapidPro. It does this through direct connections to the databases, for performance as well as customizability (The RapidPro API doesn't allow setting the contact UUID)


erikh360 (approved):
wow, lgtm!

PR #285, opened 2020-01-29
Feature/mcmr 43 add pmtct registration

OMosimege:
In this PR, we add the PMTCT registration


eventstore/models.py
```diff
@@ -197,6 +197,18 @@ class PrebirthRegistration(models.Model):
     data = JSONField(default=dict, blank=True, null=True)
 
 
+class PMTCTRegistration(models.Model):
+    id = models.UUIDField(primary_key=True, default=uuid.uuid4, editable=False)
+    contact_id = models.UUIDField()
+    device_contact_id = models.UUIDField()
+    date_of_birth = models.DateField(blank=True, null=True)
+    pmtct_risk = models.CharField(max_length=255, blank=True, default="")
```
rudigiesler: I think this should have a list of `choices`, since we know that there are only 2 choices, `high` and `normal`. It should also be required.

eventstore/test_views.py
```diff
@@ -595,6 +596,56 @@ def test_prebirth_other_passport_origin(self):
...
+            Permission.objects.get(codename="add_pmtctregistration")
+        )
+        self.client.force_authenticate(user)
+        response = self.client.post(
+            self.url,
+            {
+                "contact_id": "9e12d04c-af25-40b6-aa4f-57c72e8e3f91",
+                "device_contact_id": "d80d51cb-8a95-4588-ac74-250d739edef8",
+                "pmtct_risk": "normal",
+                "date_of_birth": "1990-02-03",
```
rudigiesler, line 629: We should also have a test to confirm that `date_of_birth` can be `null` and the request will be successful, since it's not a required field.

eventstore/models.py
OMosimege: forgot to change this, thanks

eventstore/models.py
```diff
@@ -197,6 +197,22 @@ class PrebirthRegistration(models.Model):
     data = JSONField(default=dict, blank=True, null=True)
 
 
+class PMTCTRegistration(models.Model):
+    NORMAL = "Normal"
+    HIGH = "High"
```
rudigiesler: I think these machine values should be `normal` and `high`, since that's what we use everywhere. We can keep the human values as `Normal` and `High`

eventstore/test_views.py
```diff
@@ -595,6 +596,92 @@ def test_prebirth_other_passport_origin(self):
...
+                "contact_id": "9e12d04c-af25-40b6-aa4f-57c72e8e3f91",
+                "device_contact_id": "d80d51cb-8a95-4588-ac74-250d739edef8",
+                "pmtct_risk": "Normal",
+                "date_of_birth": None,
+                "source": "WhatsApp",
+            },
+            format="json",
+        )
+
+        print(response.content)
```
rudigiesler: We can remove this print

eventstore/migrations/0018_auto_20200129_1126.py
```diff
@@ -0,0 +1,18 @@
...
+class Migration(migrations.Migration):
+
+    dependencies = [("eventstore", "0017_pmtctregistration")]
+
+    operations = [
+        migrations.AlterField(
+            model_name="pmtctregistration",
+            name="pmtct_risk",
+            field=models.CharField(
+                choices=[("Normal", "Normal"), ("High", "High")], max_length=6
```
rudigiesler: We'll probably need a new migration to apply the changes made to the model. If possible, could we delete all these new migrations, and just make it a single migration for this PR

eventstore/models.py
```diff
@@ -197,6 +197,22 @@ class PrebirthRegistration(models.Model):
     data = JSONField(default=dict, blank=True, null=True)
 
 
+class PMTCTRegistration(models.Model):
+    normal = "Normal"
+    high = "High"
```
rudigiesler: It's django convention to have the variable names uppercase, as they're considered magic numbers. What we want is the string values to be lowercase, e.g.
```
NORMAL = "normal"
RISK_TYPES = ((NORMAL, "Normal"),)
```
That will give a machine value of `normal`, and a human value of `Normal`

rudigiesler (approved):

PR #287, opened 2020-01-30
Change from referencing tel urn to whatsapp urn

rudigiesler:
With Turn handling SMS, now we don't have any tel channels in RapidPro anymore

### DIFF
--- a/eventstore/whatsapp_actions.py
+++ b/eventstore/whatsapp_actions.py
@@ -19,6 +19,7 @@ def handle_operator_message(message):
     Triggers all the tasks that need to run for an outbound message from an operator
     """
     whatsapp_contact_id = message.data["_vnd"]["v1"]["chat"]["owner"]
+    # This should be in "+27xxxxxxxxx" format, but just in case it isn't
     msisdn = normalise_msisdn(whatsapp_contact_id)
     return chain(
         get_engage_inbound_and_reply.s(),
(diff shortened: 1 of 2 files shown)


eventstore/whatsapp_actions.py
```diff
@@ -23,7 +23,8 @@ def handle_operator_message(message):
     return chain(
         get_engage_inbound_and_reply.s(),
         async_create_flow_start.s(
-            flow=settings.RAPIDPRO_OPERATOR_REPLY_FLOW, urns=[f"tel:{msisdn}"]
+            flow=settings.RAPIDPRO_OPERATOR_REPLY_FLOW,
+            urns=[f"whatsapp:{msisdn.lstrip('+')}"],
```
erikh360, line 28: Can't we just use the `whatsapp_contact_id` from above?

erikh360 (commented):
just the one comment

eventstore/whatsapp_actions.py
rudigiesler: I did a manual request, looks like `message._vnd.v1.chat.owner` is in the `+27820001001` format, so we would still have to strip the `+`. Using the normalised `msisdn` means that if this changes in the future, we'll still be okay

erikh360: Awesome!

erikh360 (approved):

PR #288, opened 2020-01-30
Handle inbound messages with EDD labels

DevChima:


erikh360:
reworked it here: https://github.com/praekeltfoundation/ndoh-hub/pull/289

PR #289, opened 2020-01-30
Handle inbound with EDD label

erikh360:


ndoh_hub/settings.py
```diff
@@ -307,3 +307,4 @@
 if ENABLE_EVENTSTORE_WHATSAPP_ACTIONS:
     RAPIDPRO_OPERATOR_REPLY_FLOW = env.str("RAPIDPRO_OPERATOR_REPLY_FLOW")
     RAPIDPRO_UNSENT_EVENT_FLOW = env.str("RAPIDPRO_UNSENT_EVENT_FLOW")
+    RAPIDPRO_EDD_LABEL_FLOW = env.str("RAPIDPRO_EDD_LABEL_FLOW")
```
erikh360, line 310: This env will need to be configured when you deploy.

DevChima (commented):
Great

ndoh_hub/settings.py
DevChima: @erikh360 Where do I configure that variable. In the sandbox Turn webhook?

DevChima: Seen. Thanks Erik

PR #290, opened 2020-02-03
Handle message expired event

erikh360:


eventstore/tasks.py
```diff
@@ -14,6 +21,10 @@
 from ndoh_hub.utils import rapidpro
 
 
+def get_utc_now():
+    return datetime.utcnow()
```
rudigiesler: We should probably work with timezone aware timestamps instead, ie. `datetime.now(tz=pytz.utc)`, to avoid timezone issues when comparing timestamps

eventstore/tasks.py
```diff
@@ -119,3 +130,125 @@ def delete_rapidpro_contact_by_uuid(contact_uuid):
...
+)
+def update_rapidpro_contact_error_timestamp(context):
+    if "msisdn" not in context:
+        return
+
+    msisdn = context["msisdn"]
+    rapidpro.update_contact(
+        f"whatsapp:{msisdn}",
+        fields={
+            "whatsapp_undelivered_timestamp": get_utc_now().strftime(
```
rudigiesler: We can probably just use `get_utc_now().isoformat()` to get the ISO formatted datetime string.

eventstore/tasks.py
```diff
@@ -119,3 +130,125 @@ def delete_rapidpro_contact_by_uuid(contact_uuid):
...
+    time_limit=15,
+)
+def get_rapidpro_contact_by_urn(urn):
+    if not urn:
+        return
+    return rapidpro.get_contacts(urn=urn).first(retry_on_rate_exceed=True).serialize()
+
+
+@app.task(
+    autoretry_for=(RequestException, SoftTimeLimitExceeded),
```
rudigiesler: We're not making any http requests in this method, so it doesn't make sense to autoretry for HTTP methods. It also probably doesn't make sense to retry with backoff either, since we're not contacting external services. `acks_late` still makes sense though (for lost workers), as well time timeouts.

eventstore/test_whatsapp_actions.py
```diff
@@ -239,8 +244,222 @@ def test_handle_whatsapp_hsm_error_unsent_disabled(self):
...
+            json.loads(turn_call.request.body),
+            {
+                "preview_url": False,
+                "recipient_type": "individual",
+                "to": "27820001001",
+                "type": "text",
+                "text": {
+                    "body": "Sibona ukuthi imiyalezo yeWhatsApp yakho yeMomConnect "
+                    "ayifikanga. Ukuthola imiyalezo nge-SMS, phendula uthi "
+                    "\u201cSMS\u201d (kusebenza izindleko ezijwayelekile)."
```
rudigiesler, line 343: These probably shouldn't be curly quotes, but that's a translation file issue, not a code issue here.

eventstore/whatsapp_actions.py
```diff
@@ -70,8 +74,7 @@ def handle_event(event):
 
 
 def handle_whatsapp_message_expired_error(event):
-    # TODO: handle whatsapp timeout system event
-    pass
+    handle_whatsapp_delivery_error(f"whatsapp:{event.recipient_id}")
```
rudigiesler: This should probably be called with `.delay`, so that it happens in the background, not in the request.

rudigiesler (approved):

PR #292, opened 2020-02-07
Admin command for migrating Registrations and Changes

rudigiesler:
This PR contains an admin command for migrating the old Registrations and Changes to the new event store models.

This is so that we can have all historical data in one place, and that we can eventually remove those older models.


OMosimege (approved):
👍 Looks good

PR #296, opened 2020-02-20
Feature/theshizz 86 change edd baby dob

DevChima:
This PR saves the new EDD and baby dob values to the event store. We've added a new feature which allows mothers to make changes to their EDD and baby dob so we need to save the changes.


eventstore/management/commands/migrate_to_eventstore.py
```diff
@@ -266,6 +268,40 @@ def handle_languageswitch_change(self, change):
...
+            defaults={
+                "contact_id": change.registrant_id,
+                "source": change.source.name,
+                "old_baby_dob": old_baby_dob,
+                "new_baby_dob": new_baby_dob,
+                "timestamp": change.created_at,
+                "created_by": change.source.user.username,
+                "data": data,
+            },
+        )
```
rudigiesler: EDD and Baby DoB changes are not something that we currently allow on Seed, so we don't need to migrate them. If there was something for Seed that we needed to migrate here, we would need to add it to the `handle` function, otherwise these functions will never be called.

ndoh_hub/urls.py
```diff
@@ -42,6 +44,8 @@
 v2router.register("postbirthregistrations", PostbirthRegistrationViewSet)
 v2router.register("researchoptins", ResearchOptinSwitchViewSet)
 v2router.register("messages", MessagesViewSet, basename="messages")
+v2router.register("eddeswitches", EddSwitchViewSet)
```
rudigiesler: Should this rather be `eddswitches`, instead of `eddeswitches`?

DevChima: Yes. Correcting that.

rudigiesler (approved):

PR #297, opened 2020-02-21
Add script for uploading whatsapp templates from csv

rudigiesler:


scripts/upload_whatsapp_templates/upload_whatsapp_templates.py
```diff
@@ -0,0 +1,131 @@
...
+from csv import DictReader
+from urllib.parse import urljoin
+
+import attr
+import requests
+
+
+def length_limit(limit):
+    def validate_length(instance, attribute, value):
+        if len(value) > 1024:
```
erikh360: This should probably use the limit var

erikh360: I also see instance is not being used but it's probably the signature required by the validator.

erikh360 (commented):
Just the one small comment but looks good otherwise

scripts/upload_whatsapp_templates/upload_whatsapp_templates.py
rudigiesler: Whoops, good catch. Yeah it's required by the validator.

erikh360 (approved):
awesome 🍰

PR #298, opened 2020-02-21
Feature/mcmr 53 opt out user after n delivery failures

OMosimege:
In this PR, we:
see if we have n consecutive SMS failures for an identity, then we opt them out


eventstore/models.py
```diff
@@ -107,6 +107,14 @@ class MSISDNSwitch(models.Model):
     data = JSONField(default=dict, blank=True, null=True)
 
 
+class DeliveryFailures(models.Model):
+    id = models.UUIDField(primary_key=True, default=uuid.uuid4, editable=False)
+    contact_id = models.CharField(max_length=255, blank=True)
```
rudigiesler: I think we can make the `contact_id` the primary key here, and remove the `id` field, because each contact should only have 1 of these to keep track of the delivery failures for that contact, so it is a unique identifier.

eventstore/test_whatsapp_actions.py
```diff
@@ -209,6 +209,28 @@ def test_delivery_failed_error(self):
...
+        DeliveryFailures.objects.create(number_of_failures=5, contact_id="27820001001")
+
+        with patch("eventstore.tasks.rapidpro") as p:
+            handle_whatsapp_delivery_error(event)
+
+        p.create_flow_start.assert_called_once_with(
+            extra={"optout_reason": "sms_failure", "timestamp": 1518694700},
+            flow="test-flow-uuid",
+            urns=["whatsapp:27820001001"],
+        )
```
rudigiesler, line 269: I think we also need a test for if there is no `DeliveryFailures` object in the database. In that case, then the `create_flow_start` should not be called, but a new `DeliveryFailures` object should be created in the database, with a `number_of_failures` equal to `1`

rudigiesler: We also need to create another test, for if the event is a successful delivery event, then the `number_of_failures` on the `DeliveryFailures` needs to be reset to 0.

eventstore/models.py
```diff
@@ -107,6 +107,14 @@ class MSISDNSwitch(models.Model):
     data = JSONField(default=dict, blank=True, null=True)
 
 
+class DeliveryFailures(models.Model):
```
rudigiesler: It is a Django best practice to name the models in singular form, ie. `DeliveryFailure`.

eventstore/models.py
```diff
@@ -107,6 +107,14 @@ class MSISDNSwitch(models.Model):
...
+    id = models.UUIDField(primary_key=True, default=uuid.uuid4, editable=False)
+    contact_id = models.CharField(max_length=255, blank=True)
+    number_of_failures = models.IntegerField(null=False, default=0)
+    timestamp = models.DateTimeField(default=timezone.now)
+    data = JSONField(default=dict, blank=True, null=True)
```
rudigiesler: I don't think that we need the `timestamp` or `data` fields in this model. This is not a model that we need to have a history for, the history is stored in the `Event` model. This is just to keep track of the current state of things.

eventstore/whatsapp_actions.py
```diff
@@ -74,7 +75,22 @@ def handle_event(event):
 
 
 def handle_whatsapp_delivery_error(event):
-    async_handle_whatsapp_delivery_error.delay(f"whatsapp:{event.recipient_id}")
+    if not event.fallback_channel:
+        async_handle_whatsapp_delivery_error.delay(f"whatsapp:{event.recipient_id}")
+    else:
```
rudigiesler, line 107: This method's name is `handle_whatsapp_delivery_error`, so it doesn't make sense that we're processing SMS/fallback channel events in here. I would make another function, `handle_fallback_delivery_error`, and call it from the `handle_event` method.

Also, this method is only called if `event.is_whatsapp_failed_delivery_event` is `True`, which won't be the case for fallback channel messages, so there's no way to get to this code if we're calling the `handle_event` function.

eventstore/models.py
rudigiesler: I think we need a `primary_key=True` here, to ensure that the contact_id is a primary key (otherwise it is possible to have duplicates)

eventstore/test_whatsapp_actions.py
```diff
@@ -244,6 +245,61 @@ def test_delivery_failed_error(self):
...
+        """
+        event = Event.objects.create()
+        event.fallback_channel = True
+        event.recipient_id = "27820001001"
+        event.timestamp = datetime.datetime(2018, 2, 15, 11, 38, 20, tzinfo=UTC)
+
+        with patch("eventstore.tasks.rapidpro") as p:
+            handle_fallback_delivery_error(event)
+
+        p.create_flow_start.assert_not_called()
```
rudigiesler, line 305: This should not only not call the flow, but should also reset the `number_of_failures` to `0`

eventstore/whatsapp_actions.py
```diff
@@ -78,6 +79,30 @@ def handle_event(event):
     if event.is_hsm_error:
         handle_whatsapp_hsm_error(event)
 
+    if event.fallback_channel is True:
+        handle_fallback_delivery_error(event)
+
+
+def handle_fallback_delivery_error(event):
```
rudigiesler: I would change this to `handle_fallback_event`, since it's not only error events going here, but also success events.

eventstore/whatsapp_actions.py
```diff
@@ -78,6 +79,30 @@ def handle_event(event):
...
+    try:
+        df = DeliveryFailure.objects.get(contact_id=event.recipient_id)
+        df.number_of_failures += 1
```
rudigiesler: This might cause a race condition when two processes are trying to increment this at the same time. Using `F` expressions, we can delegate this to the database instead, ie.
```
df.number_of_failures = F("number_of_failures") + 1
df.save(update_fields="number_of_failures")
df.refresh_from_db()
```

eventstore/whatsapp_actions.py
```diff
@@ -78,6 +79,30 @@ def handle_event(event):
...
+                    "optout_reason": "sms_failure",
+                    "timestamp": event.timestamp.timestamp(),
+                },
+                flow=settings.RAPIDPRO_OPTOUT_FLOW,
+                urns=[f"whatsapp:{event.recipient_id}"],
+            )
+    except DeliveryFailure.DoesNotExist:
+        df = DeliveryFailure(contact_id=event.recipient_id, number_of_failures=0)
+        df.number_of_failures += 1
+        df.save()
```
rudigiesler, line 109: We could just create this with a `number_of_failures` of `1` in the first place, ie.
```
DeliveryFailure.objects.create(contact_id=event.recipient_id, number_of_failures=1)
```

eventstore/migrations/0021_auto_20200302_1121.py
```diff
@@ -0,0 +1,33 @@
...
+                choices=[
+                    ("sent", "sent"),
+                    ("delivered", "delivered"),
+                    ("read", "read"),
+                    ("failed", "failed"),
+                ],
+                max_length=255,
+            ),
+        ),
+    ]
```
rudigiesler: Can we combine these two migrations into a single migration? (By deleting both, and then running `makemigrations` again)

eventstore/models.py
```diff
@@ -107,6 +107,11 @@ class MSISDNSwitch(models.Model):
     data = JSONField(default=dict, blank=True, null=True)
 
 
+class DeliveryFailure(models.Model):
+    contact_id = models.CharField(primary_key=True, max_length=255, blank=True)
```
rudigiesler: I don't think we should have `blank=True` for a primary key

eventstore/test_whatsapp_actions.py
```diff
@@ -244,6 +245,66 @@ def test_delivery_failed_error(self):
...
 
+    @override_settings(RAPIDPRO_OPTOUT_FLOW="test-flow-uuid")
+    def test_fallback_channel_delivery_failure_error(self):
+        """
+        If the event is of type Failed, and uses the fallback channel,
+        then it should trigger the message delivery failed action
+        """
+        event = Event.objects.create()
+        event.fallback_channel = True
+        event.status = "failed"
```
rudigiesler: We can use `Event.FAILED` here instead, to avoid this string.

eventstore/test_whatsapp_actions.py
```diff
@@ -244,6 +245,66 @@ def test_delivery_failed_error(self):
...
+        event.status = "read"
+        event.recipient_id = "27820001001"
+        event.timestamp = datetime.datetime(2018, 2, 15, 11, 38, 20, tzinfo=UTC)
+
+        with patch("eventstore.tasks.rapidpro") as p:
+            handle_fallback_event(event)
+
+        p.create_flow_start.assert_not_called()
+        df = DeliveryFailure.objects.get(contact_id="27820001001")
+        self.assertEqual(df.number_of_failures, 0)
```
rudigiesler, line 307: I think we need two tests for the success case here.
1. This one, that checks what happens if there is no existing DeliveryFailure
2. One that checks that if there's an existing DeliveryFailure with a number_of_failures, then the number_of_failures gets reset to 0.

rudigiesler (approved):

PR #299, opened 2020-02-25
Export message content to RapidPro campaigns

rudigiesler:
There aren't any unit tests for this, didn't think it was worth the effort, since this is a thing that's going to only be run once, and then edited in the UI in the future


OMosimege (approved):
👍

PR #302, opened 2020-03-03
added more fields for optout flow

OMosimege:


eventstore/whatsapp_actions.py
```diff
@@ -95,8 +95,12 @@ def handle_fallback_event(event):
             if df.number_of_failures >= 5:
                 async_create_flow_start.delay(
                     extra={
-                        "optout_reason": "sms_failure",
+                        "optout_reason": OptOut.SMS_FAILURE_REASON,
                         "timestamp": event.timestamp.timestamp(),
+                        "babyloss_subscription": False,
+                        "delete_info_for_babyloss": False,
+                        "delete_info_consent": False,
```
rudigiesler: I think in the USSD line we have these as the strings `"FALSE"`, not sure if it makes a difference to RapidPro, but we should probably keep it the same.

rudigiesler (commented):
Just one small thing

rudigiesler (approved):

PR #305, opened 2020-03-04
changed actions for different event statuses

OMosimege:


eventstore/whatsapp_actions.py
```diff
@@ -85,32 +85,37 @@ def handle_event(event):
...
-                        "timestamp": event.timestamp.timestamp(),
-                        "babyloss_subscription": "FALSE",
-                        "delete_info_for_babyloss": "FALSE",
-                        "delete_info_consent": "FALSE",
-                        "source": "System",
-                    },
-                    flow=settings.RAPIDPRO_OPTOUT_FLOW,
-                    urns=[f"whatsapp:{event.recipient_id}"],
+    if event.status != Event.SENT:
+        if event.status == Event.FAILED:
```
rudigiesler: This nested if is a bit difficult to follow.

I think it will be clearer if we do something like:
```
if event.status == Event.FAILED:
  ...
else if event.status == Event.DELIVERED or event.status == Event.READ:
  ...
```

rudigiesler (commented):
Just one comment, looks good otherwise

eventstore/whatsapp_actions.py
OMosimege: is it possible to do an `elif` without an `else`? because since `Event.SENT` is a [do nothing], I wouldnt need the else

rudigiesler: Yes, that's supposed to be an `elif` instead of the `else if` that I typed

rudigiesler: And yes, it is possible to not have an `else` at the end.

OMosimege: not really possible, but is it okay to do it

rudigiesler (approved):

PR #311, opened 2020-04-02
Add table for covid 19 triages

rudigiesler:
This will allow us to store the results from the USSD line


OMosimege (approved):
Looks great 👍

PR #312, opened 2020-04-06
Add CDU address update endpoint

erikh360:


eventstore/migrations/0022_cdu_address_update.py
```diff
@@ -0,0 +1,72 @@
...
+                ("date_of_birth", models.DateField(blank=True, null=True)),
+                ("folder_number", models.CharField(max_length=255)),
+                ("municipality", models.CharField(max_length=255)),
+                ("city", models.CharField(max_length=255)),
+                ("suburb", models.CharField(max_length=255)),
+                ("street_name", models.CharField(max_length=255)),
+                ("street_number", models.CharField(max_length=255)),
+            ],
+        ),
+        migrations.AlterField(
```
erikh360, line 54: Seems to be from previous PR

rudigiesler: Hmm that's weird.. I wonder if we need to `sort` this field, so that it's in the same order every time.

rudigiesler (approved):

PR #318, opened 2020-04-09
SFTP and delete CDU addres updates

DevChima:


eventstore/tasks.py
```diff
@@ -319,6 +323,76 @@ def update_rapidpro_contact_error_timestamp(context):
...
+        "city",
+        "suburb",
+        "street_name",
+        "street_number",
+    ]
+    data = []
+    ids_to_delete = []
+    for address_update in CDUAddressUpdate.objects.all():
+        ids_to_delete.append(address_update.id)
+        row = ",".join([str(getattr(address_update, header)) for header in Headers])
```
rudigiesler, line 355: This doesn't do the correct escaping for a valid CSV. Could we rather use the built-in python CSV library? https://docs.python.org/3/library/csv.html

eventstore/tasks.py
```diff
@@ -319,6 +323,76 @@ def update_rapidpro_contact_error_timestamp(context):
...
+        data.append(row)
+
+    # Write to csv
+    with open("./CDU_address.csv", "w", newline="") as csv_file:
```
rudigiesler, line 359: There is a `tempfile` library in python we can use to create a temporary file to write to, instead of relying on a static filename, where we might not have write access to.

eventstore/tasks.py
```diff
@@ -319,6 +323,76 @@ def update_rapidpro_contact_error_timestamp(context):
...
+        "district",
+        "municipality",
+        "city",
+        "suburb",
+        "street_name",
+        "street_number",
+    ]
+    data = []
+    ids_to_delete = []
+    for address_update in CDUAddressUpdate.objects.all():
```
rudigiesler, line 353: We can use `.iterator()` here to avoid loading all of the objects into memory.

eventstore/tasks.py
```diff
@@ -319,6 +323,76 @@ def update_rapidpro_contact_error_timestamp(context):
...
+    # Write to csv
+    with open("./CDU_address.csv", "w", newline="") as csv_file:
+        writer = csv.writer(csv_file)
+        writer.writerow(Headers)
+        for row in data:
+            writer.writerow(row.replace('"', "").split(","))
+
+    # Start FTP Session
+    cnopts = pysftp.CnOpts()
+    cnopts.hostkeys = None
```
rudigiesler, line 367: This will cause us to be susceptible to man-in-the-middle attacks. Can we set the key for the server in the settings file, and use it here?

That will mean that it will break if the server key ever changes, but that's exactly what we want, because that means that we're no longer talking to the same server.

There's a nice guide here to setting the key from a string: https://stackoverflow.com/questions/38939454/verify-host-key-with-pysftp/43389508#43389508

eventstore/tasks.py
```diff
@@ -319,6 +323,76 @@ def update_rapidpro_contact_error_timestamp(context):
...
+    retry_backoff=True,
+    max_retries=15,
+    acks_late=True,
+    soft_time_limit=10,
+    time_limit=15,
+)
+def read_data(data):
+    with open(data, "r") as f:
+        data = [row for row in csv.reader(f.read().splitlines())]
+        return data
```
rudigiesler, line 393: We're only using this task in the tests, what is it for?

eventstore/tasks.py
```diff
@@ -319,6 +323,76 @@ def update_rapidpro_contact_error_timestamp(context):
     )
 
 
+@app.task(
+    autoretry_for=(RequestException, ConnectionException),
```
rudigiesler, line 327: We're not making any HTTP requests here, so these are probably not useful exceptions to retry on. We should look at what exceptions `pysftp` throws for things like connection errors, that we could retry on.

eventstore/test_tasks.py
```diff
@@ -0,0 +1,102 @@
...
+            folder_number="2534567",
+            district="Lagos",
+            municipality="Lagos",
+            city="Lagos",
+            suburb="Lekki",
+            street_name="Igbo Efun",
+            street_number="10067",
+        )
+
+        upload_and_delete_cdu_address_updates(self)
```
rudigiesler, line 41: We need to have some assertions here to ensure that the file was created and uploaded properly.

eventstore/tasks.py
DevChima: Thanks. I had seen that and was thinking the same thing but wouldn't that mean that we have to get their public key and use that as the value of cnopts.hostkeys

eventstore/tasks.py
DevChima: One of them is pysftp.ConnectionException. https://pysftp.readthedocs.io/en/release_0.2.8/pysftp.html

eventstore/tasks.py
rudigiesler: Yes, we would need their public key. We could get it using `ssh-keyscan`

rudigiesler:
Closing this, since there haven't been any more updates, it doesn't seem to be required anymore

PR #319, opened 2020-04-14
Return 200 for duplicate entries

rudigiesler:
We don't want to return 400, because then it's an error that gets retried. We just want to return a 200, so that the client knows that we already have this, and can mark this as complete


erikh360 (approved):
Nice!

PR #322, opened 2020-04-30
Feature/momza 1378 add field to send sms to stuck reg identities

OMosimege:
In This PR, is a script to add a field to contacts stuck in consent state in the clinic registration


scripts/add_field_for_stuck_registrations/add_field_to_stuck_registration_contacts.py
```diff
@@ -0,0 +1,31 @@
...
+        process(row)
+
+
+def main():
+    parser = argparse.ArgumentParser(
+        description="Add field to stuck contacts consent state"
+    )
+    args = parser.parse_args()
+    loop = asyncio.get_event_loop()
+    loop.run_until_complete(process_csv(args))
```
rudigiesler: By using the rapidpro sync client, we're mixing the sync and async things here, and getting no benefit from the async, so I think we should just remove all the async and call `process_csv` directly.

scripts/add_field_for_stuck_registrations/README.md
```diff
@@ -0,0 +1,11 @@
+Add Field For Contacts Stuck In The Registration Consent State
+----------------------
+The add_field_to_stuck_registration_contacts.py script takes in a CSV in stdin,
+and outputs a CSV on stdout. The input CSV is contacts in the stuck state
```
rudigiesler: It doesn't seem to output anything on stdout

scripts/add_field_for_stuck_registrations/README.md
```diff
@@ -0,0 +1,11 @@
...
+since the implementation on asking for consent in a clinic prebirth & postbirth
+registration.
+The output CSV contains the same data as the input CSV, but with and additional
+field that notes that an sms should be sent to that contact to complete their
+registration.
```
rudigiesler: This is not true. What it does is call the RapidPro API to update the `stuck_in_consent_state` field on their contact

scripts/add_field_for_stuck_registrations/add_field_to_stuck_registration_contacts.py
```diff
@@ -0,0 +1,31 @@
...
+import csv
+import sys
+
+from ndoh_hub.utils import rapidpro
+
+
+def process(contact_id):
+    contact = rapidpro.get_contacts(uuid=contact_id).first()
+    params = {"stuck_in_consent_state": "True"}
+    rapidpro.update_contact(contact, fields=params)
```
rudigiesler: The first argument to `update_contact` is the UUID or the URN of the contact to update: https://github.com/rapidpro/rapidpro-python/blob/master/temba_client/v2/__init__.py#L461-L476 .

So we don't need to first do a `get_contacts`, we can just call `update_contact` with the UUID.

scripts/add_field_for_stuck_registrations/QA_script.csv
```diff
@@ -0,0 +1,4 @@
+64786f88-03f5-4ffd-96eb-e54ae9122d21
+f80155d1-11f1-4f69-aa21-77d5907e387e
+e2770fb4-fd91-43b9-a343-74d3ee15d19d
+7f4620b3-9cc8-4209-bd68-9f0d7c243060
```
rudigiesler: We probably don't need to add this file of UUIDs, rather make it clear in the README what kind of input this script wants.

scripts/add_field_for_stuck_registrations/README.md
rudigiesler: Also, we should probably rather mention `The input CSV is the UUID's of contacts in the stuck state`, so that it's clear what the content of the input should be.

scripts/add_field_for_stuck_registrations/QA_script.csv
OMosimege: okay, will remove it from the folder

scripts/add_field_for_stuck_registrations/add_field_to_stuck_registration_contacts.py
OMosimege: okay

scripts/add_field_for_stuck_registrations/README.md
OMosimege: I forgot to make changes to the readme, thanks for pointing this out

scripts/add_field_for_stuck_registrations/README.md
OMosimege: I forgot to make changes to the readme, thanks for pointing this out

scripts/add_field_for_stuck_registrations/add_field_to_stuck_registration_contacts.py
OMosimege: Okay, thank you

scripts/add_field_for_stuck_registrations/add_field_to_stuck_registration_contacts.py
```diff
@@ -0,0 +1,28 @@
...
+
+from ndoh_hub.utils import rapidpro
+
+
+def process(contact_id):
+    params = {"stuck_in_consent_state": "True"}
+    rapidpro.update_contact(contact_id, fields=params)
+
+
+async def process_csv(args):
```
rudigiesler, line 13: This can just be `def process_csv(args):`

OMosimege: thanks for picking this up

PR #324, opened 2020-05-07
Add new fields and API for confirmed contacts

rudigiesler:
This PR modifies existing tests to ensure that what we return on the existing API stays the same.

It then adds a new endpoint, that has the new fields


erikh360 (approved):
LGTM! Nice!

PR #325, opened 2020-05-21
Script to fix rapidpro urns

erikh360:


scripts/migrate_to_rapidpro/fix_contact_urns.py
```diff
@@ -0,0 +1,60 @@
...
+CORRECT_ORG_ID = 5
+INCORRECT_ORG_ID = 1
+
+DB = {"dbname": "rapidpro", "user": "postgres", "port": 5444, "host": "127.0.0.1"}
+
+if __name__ == "__main__":
+    conn = psycopg2.connect(**DB)
+    update_conn = psycopg2.connect(**DB)
+
+    cursor = conn.cursor()
```
rudigiesler: I think we probably want this to be a named cursor, so that it's a server side cursor, so that we don't load all the contact urns into memory: https://www.psycopg.org/docs/usage.html#server-side-cursors

rudigiesler (commented):

rudigiesler (approved):

PR #326, opened 2020-05-22
Add functionality to the API for healthcheck returning users

rudigiesler:
This allow us to POST incomplete records for healthchecks for returning users, and have it automatically filled out


miltontony:
👍 Happy with this - just worried on whether we'll be able to use this without the index.

miltontony (approved):

PR #328, opened 2020-05-27
Fix duplicate rapidpro contacts script

erikh360:


rudigiesler (approved):
:+1: The only issue I can see is if both the original contact, and the duplicate contact, both have URNs attached. I'm not sure if that's a possibility. But if it is, then this script will fail, rather than corrupt data, so happy with it.

erikh360:
@rudigiesler thanks! yes, the script handles that though, I've ran it over the data without problems.

PR #329, opened 2020-05-27
Add sync fallback channel script

erikh360:


scripts/migrate_to_rapidpro/sync_fallback_channel.py
```diff
@@ -0,0 +1,118 @@
...
+            contacts_contacturn.path,
+            contacts_contact.id,
+            contacts_contact.fields
+        FROM contacts_contacturn, contacts_contact
+        WHERE contacts_contacturn.org_id = 5
+            AND contacts_contacturn.contact_id  = contacts_contact.id
+            AND contacts_contact.org_id = 5
+            AND contacts_contacturn.scheme = 'whatsapp'
+            AND fields is not null
+            -- AND contacts_contact.id > 999
```
erikh360, line 69: Incase it fails and we need to resume

scripts/migrate_to_rapidpro/sync_fallback_channel.py
```diff
@@ -0,0 +1,118 @@
...
+                contact_details["is_fallback_active"] = False
+            elif preferred_channel == "SMS" and is_fallback_active is not True:
+                contact_details["is_fallback_active"] = True
+            else:
+                update = False
+
+            if update:
+                response = update_turn_contact_details(path, contact_details)
+
+                if response.status_code != 200:
```
erikh360: Could get rate limited here, I then print out the contact id to resume.

I did test it against a turn running locally with the same rate limit and it was happy.

scripts/migrate_to_rapidpro/sync_fallback_channel.py
```diff
@@ -0,0 +1,123 @@
...
+TURN_TOKEN = "eyJhbGciOiJIUzUxMiIsInR5cCI6IkpXVCJ9.eyJhdWQiOiJFbmdhZ2VkIiwiZXhwIjoxNjIyMTI0MjUxLCJpYXQiOjE1OTA1ODgyNjAsImlzcyI6IkVuZ2FnZWQiLCJqdGkiOiI4ZWNmMTYzNi1iMjI2LTQxMmEtYjNlYi1jZjI2ZGQ0M2VhYWQiLCJuYmYiOjE1OTA1ODgyNTksInN1YiI6Im51bWJlcjoyIiwidHlwIjoiYWNjZXNzIn0.9CuZHN84SIDiiSyChXWKDXeRU3Nl27s5HysWP38ck1QUihRXuaadViDVlIOXr6QRosTzEHqEo4hsGntsajjAjA"  # noqa
+
+
+def get_turn_contact_details(wa_id):
+    headers = {
+        "Authorization": f"Bearer {TURN_TOKEN}",
+        "content-type": "application/json",
+        "Accept": "application/vnd.v1+json",
+    }
+    response = requests.get(urljoin(TURN_URL, f"/v1/contacts/{wa_id}"), headers=headers)
```
rudigiesler, line 24: Should we do a `response.raise_for_status()`, to make sure that the status code is correct?

scripts/migrate_to_rapidpro/sync_fallback_channel.py
```diff
@@ -0,0 +1,123 @@
...
+    return json.loads(response.content)
```
rudigiesler: Could also use the `response.json()` shortcut here.

scripts/migrate_to_rapidpro/sync_fallback_channel.py
```diff
@@ -0,0 +1,123 @@
...
+    start, d_print = time.time(), time.time()
+    for (path, contact_id, fields) in cursor:
+        preferred_channel = fields.get(field_mapping["preferred_channel"], {}).get(
+            "text", ""
+        )
+
+        if preferred_channel in ("WhatsApp", "SMS"):
+            contact_details = get_turn_contact_details(path)
+
+            is_fallback_active = contact_details.get("contact_details")
```
rudigiesler: Shouldn't this be `contact_details.get("is_fallback_active")`?

scripts/migrate_to_rapidpro/sync_fallback_channel.py
rudigiesler: What about doing a `.raise_for_status` in the function instead, and wrapping the request here in a try/catch. That way we could easily wrap both requests, and have the same action if either of them fails

rudigiesler (approved):

PR #332, opened 2020-06-05
Only optout on 5 sms delivery errors

erikh360:


OMosimege (approved):
👍

rudigiesler (approved):
I see we're doing a `+=1`, followed by a `.save`.

So the race condition here is that we may miss incrementing the failure count. And when we get to 5 failures, we might trigger the flow more than once.

I think that's better than the alternative race condition handling, where we do an atomic increment, which might mean that we miss when the number of failures hits 5, and the flow ends up never being called.

We could put this processing in a lock on `recipient_id`, that will avoid us having race conditions in the first place, but happy with this solution also.

PR #334, opened 2020-06-09
Auto archive expired helpdesk contacts

erikh360:
Ticket: https://praekeltorg.atlassian.net/browse/WTS-137

In rapidpro when we set the wait_for_helpdesk flag we now also set helpdesk_timeout as the current date and helpdesk_message_id as the last message received.

I also created a dynamic group in rapidpro to query.

This task will run daily and check for expired contacts and close them.

There is another ticket to bulk archive old conversations too where I will populate the 2 new fields: https://praekeltorg.atlassian.net/browse/WTS-138

### DIFF
--- a/eventstore/tasks.py
+++ b/eventstore/tasks.py
@@ -30,7 +30,7 @@ from eventstore.models import (
     ResearchOptinSwitch,
 )
 from ndoh_hub.celery import app
-from ndoh_hub.utils import rapidpro
+from ndoh_hub.utils import get_today, rapidpro
 from registrations.tasks import request_to_jembi_api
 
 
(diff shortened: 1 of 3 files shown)


eventstore/tasks.py
```diff
@@ -349,3 +349,72 @@ def mark_turn_contact_healthcheck_complete(msisdn):
...
+    r.raise_for_status()
+
+
+@app.task(
+    autoretry_for=(RequestException, SoftTimeLimitExceeded, TembaHttpError),
+    retry_backoff=True,
+    max_retries=15,
+    acks_late=True,
+    soft_time_limit=10,
+    time_limit=15,
```
rudigiesler: 10 seconds is probably way too low for going through all the waiting for helpdesk contacts

eventstore/test_tasks.py
```diff
@@ -25,8 +31,174 @@ def test_send_successful(self):
...
+        tasks.handle_expired_helpdesk_contacts()
+
+        [_, rapidpro_update] = responses.calls
+        self.assertEqual(
+            json.loads(rapidpro_update.request.body),
+            {
+                "fields": {
+                    "helpdesk_timeout": None,
+                    "wait_for_helpdesk": None,
+                    "helpdesk_message_id": None,
```
rudigiesler, line 201: Just want to check, does doing a `POST` with `"field_name": null`, clear the field? I haven't tested that before.

erikh360: Yes it does clear it

erikh360:
@rudigiesler ready for another look, I've also made the schedule configurable, mostly for QA

rudigiesler (approved):

PR #342, opened 2020-06-23
Add disable sms failure optouts flag

erikh360:


eventstore/whatsapp_actions.py
```diff
@@ -85,24 +85,25 @@ def handle_event(event):
...
-                    "optout_reason": OptOut.SMS_FAILURE_REASON,
-                    "timestamp": event.timestamp.timestamp(),
-                    "babyloss_subscription": "FALSE",
-                    "delete_info_for_babyloss": "FALSE",
-                    "delete_info_consent": "FALSE",
-                    "source": "System",
-                },
-                flow=settings.RAPIDPRO_OPTOUT_FLOW,
-                urns=[f"whatsapp:{event.recipient_id}"],
+        if not settings.DISABLE_SMS_FAILURE_OPTOUTS:
```
rudigiesler: Could we rather do an
```
if settings.DISABLE_SMS_FAILURE_OPTOUTS:
    return
```
To avoid having such a deep nesting?

rudigiesler (approved):
Looks good, just one small suggestion

eventstore/whatsapp_actions.py
erikh360: good idea

PR #343, opened 2020-06-24
Reset failure count on new registration

erikh360:


eventstore/tasks.py
```diff
@@ -418,3 +419,15 @@ def handle_expired_helpdesk_contacts():
...
+@app.task(
+    autoretry_for=(SoftTimeLimitExceeded,),
+    retry_backoff=False,
+    max_retries=1,
+    acks_late=True,
+    soft_time_limit=10,
+    time_limit=15,
+)
+def reset_delivery_failure(contact_id):
+    DeliveryFailure.objects.filter(contact_id=contact_id).update(number_of_failures=0)
```
rudigiesler: `contact_id` is the primary key, so it has an index, so I don't think we need to do this as a background task, it should be fast enough to do it in the request.

rudigiesler (commented):

erikh360:
Thanks @rudigiesler , ready for another look

rudigiesler (approved):

PR #344, opened 2020-06-25
Add rapidpro contact lookup

erikh360:
The contact_id field in the delivery failure table refers to a WhatsApp contact_id not a Rapidpro contact uuid, we need to do a lookup on Rapidpro the get the WhatsApp contact ID, moved to task so that in doesn't happen in request.

### DIFF
--- a/eventstore/tasks.py
+++ b/eventstore/tasks.py
@@ -16,6 +16,7 @@ from eventstore.models import (
     BabySwitch,
     ChannelSwitch,
     CHWRegistration,
+    DeliveryFailure,
     EddSwitch,
     Event,
     IdentificationSwitch,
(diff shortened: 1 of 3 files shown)


eventstore/tasks.py
```diff
@@ -418,3 +419,22 @@ def handle_expired_helpdesk_contacts():
...
+    retry_backoff=False,
+    max_retries=15,
+    acks_late=True,
+    soft_time_limit=10,
+    time_limit=15,
+)
+def reset_delivery_failure(contact_uuid):
+    contact = rapidpro.get_contacts(uuid=contact_uuid).first(retry_on_rate_exceed=True)
+    wa_id = None
+    for urn in contact.urns:
```
rudigiesler, line 438: Should we maybe also handle the case where `contact` is `None`?

rudigiesler (approved):
Just one small suggestion, but looks good

PR #346, opened 2020-07-07
Fix whatsapp delivery error race condition

erikh360:
WhatsApp seems to be retrying old messages that have failed then sending the failed events in batches. Multiple of these tasks gets started at the same time before we can update the timestamp on rapidpro, this will add a lock in redis to only process one of the batch.

I wasn't able to get unit tests working

### DIFF
--- a/ndoh_hub/settings.py
+++ b/ndoh_hub/settings.py
@@ -306,11 +306,7 @@ ENABLE_JEMBI_EVENTS = env.bool("ENABLE_JEMBI_EVENTS", True)
 CACHES = {
     "default": env.cache(default="locmemcache://"),
     "locmem": {"BACKEND": "django.core.cache.backends.locmem.LocMemCache"},
-    "redis": {
-        "BACKEND": "django_redis.cache.RedisCache",
-        "LOCATION": REDIS_URL,
-        "OPTIONS": {"CLIENT_CLASS": "django_redis.client.DefaultClient"},
-    },
+    "redis": env.cache("REDIS_URL"),
 }
 
 EXTERNAL_REGISTRATIONS_V2 = env.bool("EXTERNAL_REGISTRATIONS_V2", False)
--- a/ndoh_hub/settings.py
+++ b/ndoh_hub/settings.py
@@ -306,6 +306,11 @@ ENABLE_JEMBI_EVENTS = env.bool("ENABLE_JEMBI_EVENTS", True)
 CACHES = {
     "default": env.cache(default="locmemcache://"),
     "locmem": {"BACKEND": "django.core.cache.backends.locmem.LocMemCache"},
+    "redis": {
+        "BACKEND": "django_redis.cache.RedisCache",
+        "LOCATION": REDIS_URL,
+        "OPTIONS": {"CLIENT_CLASS": "django_redis.client.DefaultClient"},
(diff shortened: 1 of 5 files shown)


ndoh_hub/settings.py
```diff
@@ -306,6 +306,11 @@
 CACHES = {
     "default": env.cache(default="locmemcache://"),
     "locmem": {"BACKEND": "django.core.cache.backends.locmem.LocMemCache"},
+    "redis": {
+        "BACKEND": "django_redis.cache.RedisCache",
+        "LOCATION": REDIS_URL,
+        "OPTIONS": {"CLIENT_CLASS": "django_redis.client.DefaultClient"},
+    },
```
rudigiesler: Can we not do a `"redis": env.cache("REDIS_URL")`, or whatever the function call looks like?

rudigiesler (commented):
Looks good, I'm just not seeing where the race condition is.

We're locking the `GET` request for fetching the contact, but that's just a GET request, that's not changing anything on the contact.

Do we not rather want the lock across the entire chain?
eg. in `get_rapidpro_contact_by_urn`, we do a 
```
lock = redis.lock(key, 3600)
if not lock.aquire(blocking=False):
  return
```

And then in `update_rapidpro_contact_error_timestamp`, do a 
```
lock = redis.lock(key, 3600)
lock.release()
```

erikh360:
@rudigiesler thanks, I think race condition wasn't the right choice of words here. 

I added it to the first task in the chain so that we don't do unnecessary api calls to Rapidpro. If `get_rapidpro_contact_by_urn` returns None instead of a contact the rest of the tasks do nothing.

rudigiesler:
@erikh360 So just to understand, what we're trying to do here is just limit the concurrency of the calls, so that we're only doing a single GET request to RapidPro per MSISDN concurrently? And also we're throwing away any tasks that come in if we're currently doing a GET request?

Do the GET requests take so long that this is an issue?

erikh360:
Yes, we're limiting it to one.

It's not really because of the speed of the GET, but the task being fired 15 times for the same urn and the timestamp to limit it being on Rapidpro. So all 15 tasks gets the same timestamp from their GET requests and thinks its fine to send a sms.

So if we get a batch like this, we can do the first one which will send the sms and update the timestamp, then drop the rest.

Does that make sense?

rudigiesler:
@erikh360 Yes that makes sense, and that's what I thought we would want to be doing. But that's not what we're doing here.

We're only limiting concurrency of the GET request here, so say we have 4 workers, and get 15 tasks all at the same time, then we'll still get this being called 3-4 times, because we're only locking on the GET request.

If we lock all the way from before the GET, to the POST that updates the timestamp, then we'll drop all the rest while the first one is processing, and any tasks after that will get the new timestamp. By locking only on the GET, it means that after the GET is complete, another task can come along and do another GET, before we've done the POST to update the timestamp, so we still end up sending multiple SMSes.

erikh360:
@rudigiesler I see what you are saying thanks, I assumed the `with r.lock(key, 3600):` would hold all the way till the end. That is also probably why I couldn't get the tests to work. 🤦 

I'll fix it.

rudigiesler (approved):

PR #354, opened 2020-09-22
Add index to clinic code lookup

rudigiesler:
Clinic code endpoint is taking on average 400ms. This is quite high for fetching a single record from the database.

I think an index should help speed this up, so that we don't need to do a full table scan every time


erikh360 (approved):
Nice

PR #357, opened 2020-11-19
Add github actions

rudigiesler:
Travis is being really slow lately, looking to switch over to github actions, starting by adding github actions in parallel


erikh360 (approved):
looks good!

PR #364, opened 2021-01-15
Add momconnect import: Sync validation

rudigiesler:
This PR adds the first of three parts, the sync validation for imports for MomConnect.

The second part will be the async validation, and the third part will be the registration


rudigiesler:
This is what the django admin looks like for an import: 
image

erikh360 (approved):

PR #369, opened 2021-02-08
Feature/mcmr 155 hub cleanup

OMosimege:
In this PR is the Hub CleanUp

What we are keeping, is the following functionality:
1. registrations.views.ThirdPartyRegistration. Currently this view uses the new things if the EXTERNAL_REGISTRATIONS_V2 config flag is set. We want to remove all the old functionality for when that flag isn’t set, and just use the new functionality, and remove the flag.

2. registrations.views.JembiAppRegistration. Here too we want to remove the old functionality, and just keep the EXTERNAL_REGISTRATIONS_V2 functionality, and remove the config flag.

3. registrations.views.WhatsAppContactCheckViewSet

4. registrations.views.FacilityCheckView and registrations.views.NCFacilityCheckView 
5. registrations.views.NCSubscriptionView, we want to keep all functionality here


changes/tasks.py
```diff
@@ -198,87 +198,10 @@ def baby_switch(self, change):
         self.log.info("Starting switch to baby")
 
         self.log.info("Retrieving active subscriptions")
-        active_subs = sbm_client.get_subscriptions(
-            {"identity": change.registrant_id, "active": True}
-        )["results"]
```
rudigiesler, line 203: I'm a little bit confused as to why we're removing certain lines, changing the functionality, instead of just deleting the whole task.

eventstore/test_whatsapp_actions.py
```diff
@@ -558,9 +558,9 @@ def test_handle_whatsapp_delivery_error(self, mock_get_utc_now):
                 "to": "27820001001",
                 "type": "text",
                 "text": {
-                    "body": "Sibona ukuthi imiyalezo yeWhatsApp yakho yeMomConnect "
-                    "ayifikanga. Ukuthola imiyalezo nge-SMS, phendula uthi "
-                    "\u201cSMS\u201d (kusebenza izindleko ezijwayelekile)."
+                    "body": "We see that your MomConnect WhatsApp "
+                    "messages are not being delivered. If you would "
+                    "like to receive your messages over SMS, reply ‘SMS’."
```
rudigiesler: These are new things, we definitely don't want to be changing these tests, although I think this functionality is currently switched off, we should keep it in case we want to switch it on again. Ensuring that the message gets translated to the user's message is important.

changes/tasks.py
```diff
@@ -1458,9 +1358,9 @@ def handle_hsm_error(self, user_id, source_id, identity_uuid):
         language = identity["details"]["lang_code"].lower().replace("_", "-")
         with translation.override(language):
             text = translation.ugettext(
-                "Sorry but we can't send WhatsApp msgs to this phone. We'll "
-                "send your MomConnect msgs on SMS. To stop dial "
-                "%(optout_ussd)s, for more dial %(popi_ussd)s (Free)."
+                "Sorry we can't send WhatsApp msgs to this phone. We'll send your "
+                "MomConnect msgs on SMS. To stop dial *134*550*1#, for more dial "
+                "*134*550*7#"
```
rudigiesler: We probably don't want to change this content, or if we do, we would want to do it in a separate PR, not as part of the cleanup. This is what's creating all the changes in the PO files.

registrations/tasks.py
```diff
@@ -314,10 +312,7 @@ def create_popi_subscriptionrequest(self, registration):
             return "POPI Subscription request not created"
 
         self.log.info("Fetching messageset")
-        msgset_short_name = utils.get_messageset_short_name(
-            "popi", registration.source.authority, None
-        )
-        r = utils.get_messageset_schedule_sequence(msgset_short_name, None)
+        r = ""
```
rudigiesler, line 315: Same here, we can just delete this whole task once we're not using it, instead of changing functionality

changes/tasks.py
OMosimege: ahh I see, okay will reverse this

registrations/tasks.py
OMosimege: This is same as above, the task is needed by another `validate_implement`, I have tried removing it and the entire validate fails, so I opted for removing the part that is not needed in the task

changes/tasks.py
OMosimege: This task is needed by another `validate_implement`, I have tried removing it and the entire validate fails, so I opted for removing the part that is not needed in the task

eventstore/test_whatsapp_actions.py
OMosimege: makes sense, will also revert this change

registrations/tasks.py
rudigiesler: Ah okay, I see, but I think all of the old `validate_implement`s are no longer needed, so we can remove all of them.

OMosimege: okay, that will help a lot, thank you

rudigiesler (approved):

PR #370, opened 2021-02-09
Add actual observation names to Ada notification

rudigiesler:
We have received what the actual observation names will be from Ada, so now we can implement them


OMosimege (approved):
Great work 👍

PR #371, opened 2021-02-09
Add additional ada observations

rudigiesler:
They've supplied us with additional observations that we can use to fill in some of the optional fields


OMosimege (approved):
👍

PR #380, opened 2021-05-03
ADA link tracking

DevChima:
WIP


ada/models.py
```diff
@@ -0,0 +1,28 @@
+from django.db import models
+
+# Create your models here.
+
+
+class RedirectUrl(models.Model):
+
+    urls = models.CharField(max_length=255, blank=True, null=True)
```
rudigiesler: Shouldn't this just be `url`, since we're putting a single URL here? Also, we can probably use django's `URLField` here instead for some additional validation

ada/views.py
```diff
@@ -0,0 +1,24 @@
...
+        return HttpResponseRedirect(my_link)
+    except Exception as e:
+        return render(request, 'index.html', {'error': e})
+
+def default_page(request,pk):
+    try:
+        redirect_url = RedirectUrl.objects.get(id=pk)
+        whatsappid = request.GET.get("whatsappid")
+        return render(request, 'meta_refresh.html', {'pk':pk,'whatsappid':whatsappid})
+    except Exception as e:
```
rudigiesler: We definitely want to remove all of these before merging

ada/views.py
```diff
@@ -0,0 +1,24 @@
...
+
+
+def clickActivity(request, pk):
+
+    try:
+        redirect_url = RedirectUrl.objects.get(id=pk)
+
+        store_url_entry = RedirectUrlsEntry(urls=redirect_url)
+        store_url_entry.save()
+        my_link = f'http://symptomcheck.co.za/?whatsappid={request.GET.get("whatsappid")}'
```
rudigiesler: Shouldn't this be based on `redirect_url.urls` instead of a hardcoded value?

ada/models.py
```diff
@@ -0,0 +1,28 @@
...
+    content = models.TextField(default='This url has no copy')
```
rudigiesler: We don't seem to be using this field

template/meta_refresh.html
```diff
@@ -0,0 +1,12 @@
+<!DOCTYPE html>
+<html lang="en">
+    <meta http-equiv="refresh" content="3; url=https://hub.momconnect.co.za/confirmredirect/{{pk}}?whatsappid={{whatsappid}}" />
```
rudigiesler: I think we can go all the way to `0` for the redirect time, to redirect instantly.

rudigiesler: We also don't want this URL to be hardcoded

ada/models.py
```diff
@@ -0,0 +1,45 @@
...
+    url = models.URLField(
+        max_length=255,
+        blank=True,
+        null=True,
+        default="https://hub.momconnect.za/confirmredirect",
+    )
+    content = models.TextField(default="This entry has no copy")
+    refresh_url = models.CharField(
+        max_length=200, null=False, blank=False, default="https://hub.momconnect.za"
+    )
```
rudigiesler, line 12: For this, we're redirecting back to ourselves, so we don't need a database field for this

ada/views.py
```diff
@@ -0,0 +1,28 @@
...
+    try:
+        redirect_url = RedirectUrl.objects.get(id=pk)
+        store_url_entry = RedirectUrlsEntry(url=redirect_url)
+        store_url_entry.save()
+        destination_url = (
+            f"{redirect_url.symptom_check_url}"
+            f'/?whatsappid={request.GET.get("whatsappid")}'
+        )
+        return HttpResponseRedirect(destination_url)
+    except Exception as e:
```
rudigiesler: We don't want these generic catching of exceptions, rather catch ones we can handle properly, and let the others go into sentry so that we can know that they exist and fix the bug that's causing them

ndoh_hub/settings.py
```diff
@@ -25,15 +25,15 @@
...
+TEMPLATES_DIR = os.path.join(BASE_DIR, "template")
 # Quick-start development settings - unsuitable for production
 # See https://docs.djangoproject.com/en/1.9/howto/deployment/checklist/
 
 # SECURITY WARNING: keep the secret key used in production secret!
 SECRET_KEY = os.environ.get("SECRET_KEY", "REPLACEME")
 
 # SECURITY WARNING: don't run with debug turned on in production!
-DEBUG = os.environ.get("DEBUG", False)
+DEBUG = os.environ.get("DEBUG", True)
```
rudigiesler: We should default to False for debug.

template/meta_refresh.html
```diff
@@ -0,0 +1,12 @@
+<!DOCTYPE html>
+<html lang="en">
+    <meta http-equiv="refresh" content="0; url={{ refresh_url }}/confirmredirect/{{pk}}?whatsappid={{whatsappid}}" />
```
rudigiesler: We should rather use the url template tag for building urls in templates: https://docs.djangoproject.com/en/3.1/ref/templates/builtins/#url

ada/models.py
```diff
@@ -0,0 +1,45 @@
...
+
+class RedirectUrl(models.Model):
+
+    url = models.URLField(
+        max_length=255,
+        blank=True,
+        null=True,
+        default="https://hub.momconnect.za/confirmredirect",
+    )
+    content = models.TextField(default="This entry has no copy")
```
rudigiesler: We don't seem to be using this `content` field anywhere except for in the `__str__` method. What is it for?

ada/models.py
```diff
@@ -0,0 +1,45 @@
...
+    refresh_url = models.CharField(
+        max_length=200, null=False, blank=False, default="https://hub.momconnect.za"
+    )
+    symptom_check_url = models.CharField(
+        max_length=200, null=False, blank=False, default="http://symptomcheck.co.za"
+    )
```
rudigiesler: This is the URL that we're redirecting to. Why are we using `symptom_check_url`, a `CharField`, instead of `url`, a `UrlField`? It seems like `url` is not being used at all

ada/models.py
DevChima: Content returns the copy of the content for that URL. I can't attach an image with this comment so I'll share in Slack. It defaults to "...entry has no copy"

ada/models.py
```diff
@@ -0,0 +1,45 @@
...
+
+
+class RedirectUrl(models.Model):
+
+    url = models.URLField(
+        max_length=255,
+        blank=True,
+        null=True,
+        default="https://hub.momconnect.za/confirmredirect",
+    )
```
rudigiesler: I'm not sure what this `url` field is being used for? It seems like it's only being used for display, but that seems odd. What is this URL?

ada/models.py
```diff
@@ -0,0 +1,45 @@
...
+        if self.url:
+            return (
+                f"{self.parameter}: {self.url}/{self.id} \n"
+                f"| Clicked {self.my_counter()} times | Content: {self.content}"
+            )
+
+
+class RedirectUrlsEntry(models.Model):
+    url = models.ForeignKey(
+        RedirectUrl, on_delete=models.CASCADE, blank=True, null=True
```
rudigiesler: Do we want to allow this field to be blank or null?

ada/models.py
```diff
@@ -0,0 +1,45 @@
...
+        blank=True,
+        null=True,
+        default="https://hub.momconnect.za/confirmredirect",
+    )
+    content = models.TextField(
+        default="This entry has no copy",
+        help_text="The content of the mesage that this link was sent in",
+    )
+    symptom_check_url = models.URLField(
+        max_length=200, null=False, blank=False, default="http://symptomcheck.co.za"
```
rudigiesler: Do we want to allow this field to be blank or null? Surely we would always want to redirect somewhere?

ada/views.py
```diff
@@ -0,0 +1,33 @@
...
+from django.shortcuts import HttpResponseRedirect, render
+
+from .models import RedirectUrl, RedirectUrlsEntry
+
+
+def clickActivity(request: HttpRequest, pk: int, whatsappid: int) -> HttpResponse:
+    try:
+        redirect_url = RedirectUrl.objects.get(id=pk)
+    except (ValueError, RedirectUrl.DoesNotExist):
+        return render(request, "index.html", {"error": "HTTP Bad Request"})
```
rudigiesler: We should probably just return a 404 in this case. There's a `get_object_or_404` method that we can use to easily do that: https://docs.djangoproject.com/en/3.2/topics/http/shortcuts/#get-object-or-404

ada/views.py
```diff
@@ -0,0 +1,33 @@
...
+    store_url_entry = RedirectUrlsEntry(url=redirect_url)
+    store_url_entry.save()
+    destination_url = f"{redirect_url.symptom_check_url}" f"/?whatsappid={whatsappid}"
```
rudigiesler: It seems odd that we're adding a `/` here, what if it URL already has a `/`?

We should also use https://docs.python.org/3/library/urllib.parse.html#urllib.parse.quote to create the querystring, to ensure that it's a valid querystring.

ada/views.py
```diff
@@ -0,0 +1,33 @@
...
+    return HttpResponseRedirect(destination_url)
+
+
+def default_page(request: HttpRequest, pk: int) -> HttpResponse:
+    try:
+        whatsappid = request.GET.get("whatsappid")
```
rudigiesler: In `urls.py`, we're doing a `confirmredirect/<int:pk>/<int:whatsappid>`, so the whatsapp ID is in the URL, not in the querystring.

ada/views.py
```diff
@@ -0,0 +1,33 @@
...
+        if whatsappid is None:
+            return render(
+                request, "index.html", {"error": "404 Bad Request: Whatsappid is none"}
+            )
+    except (NameError):
```
rudigiesler: What would cause this `NameError` to happen? I can't see where this would be triggered.

rudigiesler (commented):
We'll also need unit tests for all of this

ada/models.py
rudigiesler: I see, the content is the content of the message that contained this URL, it's for us to easily see that information in the django admin

ada/tests_views.py
```diff
@@ -0,0 +1,64 @@
...
+
+from .models import RedirectUrl
+
+
+class TestViews(TestCase):
+    def test_whatsapp_url_error(self):
+        """
+        Should check that the right template is used for no whatsapp string in url
+        """
+        response = self.client.get("https://hub.momconnect.za/redirect/52?")
```
rudigiesler: We should be using `reverse` to get these URLs, not hardcoding them

ada/models.py
rudigiesler: This doesn't seem to be resolved

ada/models.py
rudigiesler: This doesn't seem to be resolved

ada/tests_views.py
```diff
@@ -0,0 +1,64 @@
...
+            time_stamp="2021-05-06 07:24:14.014990+00:00",
+        )
+
+    def tearDown(self):
+        super(AdaHookViewTests, self).tearDown()
+        self.post.delete()
+
+    def test_ada_hook_redirect_success(self):
+        response = self.client.get(
+            reverse("ada_hook_redirect", args=(self.post.id, self.post.id))
```
rudigiesler: The second argument here is the whatsapp ID, not the post ID again

ada/tests_views.py
```diff
@@ -0,0 +1,64 @@
...
+        )
+        self.assertEqual(response.status_code, 302)
+
+    def test_ada_hook_redirect_404(self):
+        response = self.client.get(reverse("ada_hook_redirect", args=("1", "1")))
```
rudigiesler: How do we know that the above created post ID is not 1? I think we should make it something that it can't possibly be, like letters, eg. `("invalidurlid", "invalidwhatsappid")`

ada/views.py
```diff
@@ -0,0 +1,34 @@
...
+    try:
+        redirect_url = RedirectUrl.objects.get(id=pk)
+    except RedirectUrl.DoesNotExist:
+        raise Http404()
+    else:
+        store_url_entry = RedirectUrlsEntry(url=redirect_url)
+        store_url_entry.save()
+        url = f"{redirect_url.symptom_check_url}" f"?whatsappid={whatsappid}"
+        parsed_url = urlparse(url)
+        destination_url = parsed_url.geturl()
```
rudigiesler: This will just make sure that if it's a valid URL, then it carries on, but if it's an invalid URL, then it raises an exception and results in a 500 error. That's not quite what we want to do here. We rather want to use https://docs.python.org/3/library/urllib.parse.html#urllib.parse.quote to quote the querystring part of the URL, and then combine the URL and the querystring to create the destination URL

ada/urls.py
```diff
@@ -0,0 +1,12 @@
+from django.urls import path
+
+from . import views
+
+urlpatterns = [
+    path(
+        "confirmredirect/<int:pk>/<int:whatsappid>",
```
rudigiesler: whatsappid is not an int, it's the rapidpro contact ID, which is a UUID, so we should allow strings here

ada/views.py
```diff
@@ -0,0 +1,34 @@
...
+
+
+def default_page(request: HttpRequest, pk: int) -> HttpResponse:
+    try:
+        whatsappid = request.GET.get("whatsappid")
+        if whatsappid is None:
+            return render(
+                request, "index.html", {"error": "404 Bad Request: Whatsappid is none"}
+            )
+    except RuntimeError:
```
rudigiesler: Why are we ignoring runtimeerrors here? That seems wrong

ada/views.py
```diff
@@ -0,0 +1,35 @@
...
+        redirect_url = RedirectUrl.objects.get(id=pk)
+    except (ValueError, RedirectUrl.DoesNotExist):
+        raise Http404()
+    else:
+        store_url_entry = RedirectUrlsEntry(url=redirect_url)
+        store_url_entry.save()
+        url = f"{redirect_url.symptom_check_url}"
+        qs = f"?whatsappid={whatsappid}"
+        parsed_url = urlparse(qs)
+        destination_url = urljoin(url, parsed_url.geturl())
```
rudigiesler: I'm not sure what these 3 lines are doing, but if we're trying to create a proper quoted querystring, then we probably want:
```
qs = urllib.parse.urlencode({"whatsappid": whatsappid})
destination_url = f"{redirect_url.symptom_check_url}?{qs}"
```

ada/views.py
```diff
@@ -0,0 +1,35 @@
...
+
+def default_page(request: HttpRequest, pk: int) -> HttpResponse:
+    try:
+        whatsappid = request.GET.get("whatsappid")
+        if whatsappid is None:
+            return render(
+                request, "index.html", {"error": "404 Bad Request: Whatsappid is none"}
+            )
+    except RuntimeError as exc:
+        raise RuntimeError("Runtime error in application") from exc
```
rudigiesler: Why are we catching a RuntimeError?

ada/urls.py
```diff
@@ -0,0 +1,17 @@
...
+    path(
+        "confirmredirect/<int:pk>/<str:whatsappid>",
+        views.clickActivity,
+        name="ada_hook_redirect",
+    ),
+    path(
+        "confirmredirect/<str:pk>/<str:whatsappid>",
+        views.clickActivity,
+        name="ada_hook_redirect",
+    ),
```
rudigiesler, line 15: This seems to be the same URL pattern duplicated

ada/tests_views.py
```diff
@@ -0,0 +1,78 @@
...
+    def test_ada_hook_redirect_success(self):
+        response = self.client.get(
+            reverse("ada_hook_redirect", args=(self.post.id, "1235"))
+        )
+        self.assertEqual(response.status_code, 302)
+
+    # Raise HTTp404 if RedirectUrl does not exist
+    def test_ada_hook_redirect_404(self):
+        response = self.client.get(
+            reverse("ada_hook_redirect", args=("1", "27789049372"))
```
rudigiesler, line 77: How do we know that `self.post.id` is not `1`? Also, is this not just a duplicate of the `test_ada_hook_redirect_404_nameError` test?

ada/tests_views.py
```diff
@@ -0,0 +1,78 @@
...
+        )
+        self.assertEqual(response.status_code, 404)
+
+    # Raise HTTp404 if ValueError
+    def test_ada_hook_redirect_404_nameError(self):
+        response = self.client.get(
+            reverse("ada_hook_redirect", args=("invalidurlid", "invalidwhatsappid"))
+        )
+        self.assertEqual(response.status_code, 404)
```
rudigiesler, line 86: These tests aren't being run, you'll need to add these test files to `setup.cfg` to have the tests running and to have code coverage reports for them.

ada/tests_views.py
DevChima: @rudigiesler This returns a 404 if the arguments are present but the RedirectUrl object does not exist. 
The other one returns  a 404 for a ValueError if say the whatsapp id is passed as a string.

ada/urls.py
DevChima: This pattern is for <str/str> rather than <int/str> so the view returns a valueError in this case

ada/views.py
DevChima: I think they might be doing the same thing just that I may have gone a longer route. 
Url is the url without the query string
qs will return query string as whatsappid=12345
Urlparse is a urllib function that parses the url into 6 components (ParseResult(scheme='http', netloc='netloc', path='/path', params='parameters', query='query=argument', fragment='fragment'))
In destination_url, it gets assembled together with .geturl() and joined to the url.

ada/models.py
```diff
@@ -0,0 +1,43 @@
...
+    def my_counter(self):
+        total_number = RedirectUrlsEntry.objects.filter(
+            symptom_check_url=self.id
+        ).count()
+        return total_number
+
+    def __str__(self):
+        if self.symptom_check_url:
+            return (
+                f"{self.parameter}: https://hub.momconnect.co.za/redirect/{self.id} \n"
```
rudigiesler: It seems odd that `hub.momconnect.co.za` is hardcoded here... In QA this will show the production URL. But we also don't have the `request` here to create a full URL. 

There is a `get_absolute_url` method on the model, which if included, django admin will create a `View on site` button with the URL. That seems like the more appropriate method to use https://docs.djangoproject.com/en/3.1/ref/models/instances/#django.db.models.Model.get_absolute_url

ada/tests_views.py
```diff
@@ -0,0 +1,95 @@
...
+        Should use the meta refresh template if url is correct
+        """
+        qs = "?whatsappid=12345"
+        url = urljoin(reverse("ada_hook", args=["1"]), qs)
+        response = self.client.get(url)
+        self.assertTemplateUsed(response, "meta_refresh.html")
+
+    def test_special_character(self):
+        """
+        Should use the meta refresh template if url is correct
```
rudigiesler: This description doesn't match up with what the test is asserting. Special characters in query strings need to be encoded with https://docs.python.org/3/library/urllib.parse.html#urllib.parse.urlencode, and if they are encoded properly, then it should just work without error

setup.cfg
```diff
@@ -3,7 +3,7 @@ exclude = */migrations/*.py,*/manage.py,ve/*
 max-line-length = 88
 
 [tool:pytest]
-python_files=registrations/test*.py registrations/**/test*.py changes/test*.py changes/**/test*.py ndoh_hub/test*.py scripts/**/test*.py eventstore/test*.py eventstore/**/test*.py
+python_files=registrations/test*.py registrations/**/test*.py changes/test*.py changes/**/test*.py ndoh_hub/test*.py scripts/**/test*.py eventstore/test*.py eventstore/**/test*.py ada/test*.py ada/**/test*.py
 addopts = --verbose --ds=ndoh_hub.testsettings --ignore=ve --cov=ndoh_hub --cov=registrations --cov=changes --cov=scripts --cov=eventstore --no-cov-on-fail
```
rudigiesler: We'll also need to add a `--cov=ada` to have test coverage report on the `ada` folder

ada/views.py
rudigiesler: It doesn't give me the same results when I try it:
```python
>>> from urllib.parse import urlparse, urlencode
>>> urlparse("?whatsappid=+1234").geturl()
'?whatsappid=+1234'
>>> urlencode({"whatsappid": "+1234"})
'whatsappid=%2B1234'
```

ada/views.py
```diff
@@ -0,0 +1,38 @@
...
+        store_url_entry.save()
+        url = f"{redirect_url.symptom_check_url}"
+        qs = f"?whatsappid={whatsappid}"
+        parsed_url = urlparse(qs)
+        destination_url = urljoin(url, parsed_url.geturl())
+        return HttpResponseRedirect(destination_url)
+
+
+def default_page(request: HttpRequest, pk: int) -> HttpResponse:
+    regex = re.compile("[@_!#$%^&*()<>?/\|}{~:]")
```
rudigiesler: These are not all possible special characters.. What if I have a \`, or a 😀 . 

I think it's better just to properly encode and forward on any special characters in the URL, rather than returning an error here.

rudigiesler (approved):

PR #381, opened 2021-05-11
HCS randomize users into study

erikh360:


eventstore/test_turn_tasks.py
```diff
@@ -0,0 +1,27 @@
+import json
+
+import responses
+from django.test import TestCase, override_settings
+
+from eventstore.turn_tasks import update_turn_contact
+
+
+class UpdateTurnContactTaskTest(TestCase):
```
erikh360, line 9: Had to move this into a separate file to avoid circular imports

rudigiesler (approved):

PR #388, opened 2021-05-19
Endpoint to receive from ADA

DevChima:


ada/tasks.py
```diff
@@ -0,0 +1,29 @@
...
+from django.conf import settings
+from requests.exceptions import RequestException
+from temba_client.exceptions import TembaHttpError
+from temba_client.v2 import TembaClient
+
+from ndoh_hub.celery import app
+
+
+rapidpro = None
+if settings.ADA_RAPIDPRO_URL and settings.ADA_RAPIDPRO_TOKEN:
```
erikh360: Does ADA have its own rapidpro org? not on momconnect?

ada/views.py
```diff
@@ -2,8 +2,23 @@
...
 from .models import RedirectUrl, RedirectUrlsEntry
+from .tasks import submit_whatsappid_to_rapidpro
+
+
+class RapidProStartFlowView(APIView):
+    permission_classes = (IsAuthenticated,)
+
+    def post(self, request: HttpRequest) -> HttpResponse:
+        whatsappid = request.query_params["whatsappid"]
+        submit_whatsappid_to_rapidpro(whatsappid)
```
erikh360: This would need to be `.delay()` so that it doesn't happen in request and rather goes to the celery queue

erikh360: ...and maybe add a test for the view with a patch on the task, then task.delay.assertCalledWith...

ndoh_hub/settings.py
```diff
@@ -51,8 +51,13 @@
     "django.contrib.staticfiles",
     # 3rd party
     "raven.contrib.django.raven_compat",
-    "rest_framework",
     "rest_framework.authtoken",
+    "rest_auth",
```
erikh360: Are all these auth things needed? Can't we do token auth?

erikh360 (commented):
few comments and questions

ada/tasks.py
DevChima: No it doesn't. I wasn't sure this was so I thought it was just a project reference thing. I'll remove

ndoh_hub/settings.py
DevChima: Token should be fine. I'll remove rest_auth and the url

ndoh_hub/settings.py
```diff
@@ -51,8 +51,13 @@
     "django.contrib.staticfiles",
     # 3rd party
     "raven.contrib.django.raven_compat",
-    "rest_framework",
     "rest_framework.authtoken",
+    "rest_framework",
+    # "django.contrib.sites",
```
erikh360: we can remove these completely

erikh360 (approved):
just the one comment

PR #390, opened 2021-05-21
Rapidpro org settings

DevChima:
Rapidpro=None


ndoh_hub/settings.py
```diff
@@ -350,7 +350,12 @@
...
 
 # ADA
+ADA_SYMPTOMCHECK_V2 = env.bool("ADA_SYMPTOMCHECK_V2", False)
 ADA_RAPIDPRO_URL = env.str("ADA_RAPIDPRO_URL", None)
 ADA_RAPIDPRO_TOKEN = env.str("ADA_RAPIDPRO_TOKEN", None)
 
 ADA_PROTOTYPE_SURVEY_FLOW_ID = env.str("ADA_PROTOTYPE_SURVEY_FLOW_ID", None)
+
+if ADA_SYMPTOMCHECK_V2:
+    RAPIDPRO_URL = env.str("RAPIDPRO_URL")
```
erikh360: are these different to the rapidpro settings on line 354?

and is it necessary to have the `ADA_SYMPTOMCHECK_V2` flag?

erikh360 (commented):
just the two questions.

also I'm a bit hesitant to merge this without fixing the tests, we're bound to miss something if the tests are failing

ada/utils.py
```diff
@@ -0,0 +1,8 @@
+from __future__ import absolute_import, division
+
+from django.conf import settings
+from temba_client.v2 import TembaClient
+
+rapidpro = None
+if settings.EXTERNAL_REGISTRATIONS_V2:
```
erikh360: I don't think we need this, maybe check if RAPIDPRO_URL and RAPIDPRO_TOKEN is available

erikh360 (approved):
just the one

ada/utils.py
DevChima: Fixed

erikh360 (approved):

PR #394, opened 2021-06-09
First script to fix auto baby switch data

erikh360:
This script will look at the eventstore and rapidpro data to determine which contacts didn't get switched over automatically.

It will output two files:
One to update the contact in Rapidpro
One to push the baby switches to the eventstore(I plan on using the output from this to submit to Jembi)

### DIFF
--- a/scripts/migrate_to_rapidpro/fix_auto_baby_switch_data.py
+++ b/scripts/migrate_to_rapidpro/fix_auto_baby_switch_data.py
@@ -15,7 +15,7 @@ RAPIDPRO_OUTPUT_FILE = "rapidpro_babyswitch_updates.csv"
 
 LIMIT = 10_000_000
 RAPIDPRO_URL = "https://rapidpro.prd.momconnect.co.za/"
-RAPIDPRO_TOKEN = os.environ['RAPIDPRO_TOKEN']
+RAPIDPRO_TOKEN = os.environ["RAPIDPRO_TOKEN"]
 
 HUB_DB_PASSWORD = os.environ["HUB_PASS"]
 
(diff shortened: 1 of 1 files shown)


scripts/migrate_to_rapidpro/fix_auto_baby_switch_data.py
```diff
@@ -0,0 +1,259 @@
...
+
+def get_babyswitches(conn):
+    babyswitches = {}
+    cursor = conn.cursor("baby_switches")
+    print("Fetching Baby Switches...")
+    cursor.execute(
+        f"""
+        select contact_id, timestamp
+        from eventstore_babyswitch
+        order by timestamp desc
```
rudigiesler: By ordering desc, and then putting it in a dictionary, we'll be overwriting and always having the oldest one per user. Don't we always want the newest one?

scripts/migrate_to_rapidpro/fix_auto_baby_switch_data.py
```diff
@@ -0,0 +1,259 @@
...
+        limit {LIMIT}
+        """
+    )  # 158680
+    total = 0
+    start, d_print = time.time(), time.time()
+    for (contact_id, timestamp) in cursor:
+        babyswitches[contact_id] = {"timestamp": timestamp}
```
rudigiesler: We can probably save a bunch of memory by just storing `timestamp`, instead of a dictionary with the timestamp in it, I can't see anywhere where we would require this to be a dict

rudigiesler (commented):
Looks good, my main question is around the ordering of the SQL queries

rudigiesler (approved):

PR #397, opened 2021-06-10
POST and start top up flow in rapidpro

DevChima:


ada/tasks.py
```diff
@@ -22,3 +26,21 @@ def submit_whatsappid_to_rapidpro(whatsappid):
             flow=settings.ADA_PROTOTYPE_SURVEY_FLOW_ID,
             urns=[f"whatsapp:{whatsappid.lstrip('+')}"],
         )
+
+
+def submit_whatsappid_to_rapidpro_topup(whatsappid):
```
erikh360: Missing the `@app.task(` decorator for these two tasks

template/topup_request.html
```diff
@@ -0,0 +1,11 @@
+<!DOCTYPE html>
+<html lang="en">
+<head>
+    <meta charset="UTF-8">
+    <meta name="viewport" content="width=device-width, initial-scale=1.0">
+    <title>Document</title>
+</head>
+<body>
+    <h1>Your topup request has been received.</h1>
```
erikh360: How is this view called? not saying its wrong, just not sure why this template is needed.

ada/tasks.py
erikh360: I'm thinking the two `submit_whatsappid_to_rapidpro*` tasks could be better named, bit confusing.

`start_prototype_survey_flow` and `start_topup_flow` is a bit more descriptive of what they do.

erikh360 (commented):
looks good but few comments and Q's

template/topup_request.html
DevChima: So the user clicks on a Whatsapp link first which then shows this template letting the contact know that their top-up request has been received and then we fire a POST to the endpoint to start the flow. So this view is just for user experience.

ada/tasks.py
DevChima: I agree.

template/topup_request.html
DevChima: https://drive.google.com/file/d/1hxLEkEwxdPtDr3pWX-DK9_YonrB3w0PJ/view?usp=sharing

ada/tasks.py
```diff
@@ -15,10 +19,44 @@
...
+    if rapidpro and settings.ADA_TOPUP_FLOW_ID:
+        return rapidpro.create_flow_start(
+            extra={},
+            flow=settings.ADA_TOPUP_FLOW_ID,
+            urns=[f"whatsapp:{whatsappid.lstrip('+')}"],
+        )
+
+
+@app.task(
+    autoretry_for=(RequestException, SoftTimeLimitExceeded, TembaHttpError),
```
erikh360: TembaHttpError is a rapidpro specific error, not possible here

DevChima: Removed

erikh360 (approved):

PR #399, opened 2021-06-24
Adds customization id query string

DevChima:
We're currently doing this redirect:
http://symptomcheck.co.za/?whatsappid=12345

But ADA now wants this: 
http://symptomcheck.co.za/?whatsappid=12345?customizationId=kh93qnNLps

Checked with ADA and url should be
http://symptomcheck.co.za/?whatsappid=12345&customizationId=kh93qnNLps


ada/views.py
```diff
@@ -33,9 +33,11 @@ def clickActivity(request: HttpRequest, pk: int, whatsappid: str) -> HttpRespons
     else:
         store_url_entry = RedirectUrlsEntry(symptom_check_url=redirect_url)
         store_url_entry.save()
+        customizationId = "kh93qnNLps"
```
erikh360: Can we add this to settings with this value as a default please?

erikh360 (commented):
Just the one comment and I think a test would be nice for this.

ada/views.py
DevChima: Alright.

ada/views.py
```diff
@@ -33,9 +33,11 @@ def clickActivity(request: HttpRequest, pk: int, whatsappid: str) -> HttpRespons
...
         url = f"{redirect_url.symptom_check_url}"
-        qs = urlencode({"whatsappid": whatsappid})
-        destination_url = f"{url}?{qs}"
+        whatsapp_qs = urlencode({"whatsappid": whatsappid})
```
erikh360: Is it not possible to urlencode this all in one step?

DevChima: Might be. Let me see

PR #400, opened 2021-06-24
ADA customization id and tests

DevChima:
We're currently doing this redirect:
http://symptomcheck.co.za/?whatsappid=12345

But ADA now wants this:
http://symptomcheck.co.za/?whatsappid=12345&customizationId=kh93qnNLps


ada/views.py
```diff
@@ -47,8 +48,9 @@ def clickActivity(request: HttpRequest, pk: int, whatsappid: str) -> HttpRespons
     else:
         store_url_entry = RedirectUrlsEntry(symptom_check_url=redirect_url)
         store_url_entry.save()
+        customizationId = settings.ADA_CUSTOMIZATION_ID
```
erikh360: small one, but why not use _ instead of camelcase

DevChima: That's fine

erikh360 (approved):

PR #401, opened 2021-06-30
add throttle env vars

codiebeulaine:


ndoh_hub/settings.py
```diff
@@ -211,8 +211,8 @@
     "DEFAULT_FILTER_BACKENDS": ("django_filters.rest_framework.DjangoFilterBackend",),
     "DEFAULT_THROTTLE_CLASSES": ["rest_framework.throttling.ScopedRateThrottle"],
     "DEFAULT_THROTTLE_RATES": {
-        "covid19triage.create": "30/second",
-        "covid19triage.list": "60/minute",
+        "covid19triage.create": os.environ.get("CREATE_THROTTLE_RATE", "30/second"),
```
erikh360: Can we add `covid19triage` in this env var name please? if we want to start throttling more endpoints its going to get confusing

erikh360 (commented):
Just the one comment for both

ndoh_hub/settings.py
codiebeulaine: yep

erikh360 (approved):
happy if you fix the lint errors

PR #403, opened 2021-07-12
Add baby dob fields to import

erikh360:


eventstore/models.py
```diff
@@ -966,6 +969,24 @@ def clean(self):
...
+            pass
+
+        if (
+            not self.edd_year
+            and not self.edd_month
+            and not self.edd_day
+            and not self.baby_dob_year
+            and not self.baby_dob_month
+            and not self.baby_dob_day
+        ):
```
rudigiesler, line 987: Should we not rather do something like
`if not any((all((self.edd_year, self.edd_month, self.edd_day)), all((self.baby_dob_year, self.baby_dob_month, self.baby_dob_day)))):`? Otherwise if we just supply `baby_dob_year`, and none of the other fields, then this validation passes.

erikh360: hmm cool, I was wondering if there was a way to do that, thanks

erikh360: The problem is when you specify a string instead of a int for a month, it also triggers with this if statement.

So I think its fine as is. If they just supply a year the other rule should fire

rudigiesler (approved):

PR #406, opened 2021-09-30
Update hcs study randomization + remove pilot c study

erikh360:
We're not doing a pilot for study C anymore so removing it.

The idea with the randomisation is that we get the minimum total per province then we check against the percentage target.

Not sure if there is a better way to do it but open to suggestions.


eventstore/models.py
```diff
@@ -668,22 +657,23 @@ def has_value(v):
                 self.data[k] = v
 
     def update_post_screening_study_arms(self, risk, source):
-        if self.age == Covid19Triage.AGE_U18:
+        if self.age == Covid19Triage.AGE_U18 and not self.province:
```
rudigiesler: I'm not sure what the change is supposed to be, just want to double check that this is correct, seems odd that they would be under 18, but some of them have a province and others don't

eventstore/models.py
```diff
@@ -709,33 +699,38 @@ def process_study_c(self, risk, source):
...
-                self.msisdn, None, None, self.hcs_study_c_pilot_arm, risk, source
-            )
+        actual_total = {
+            province["province"]: province["province__count"]
+            for province in all_provinces.annotate(models.Count("province"))
+        }.get(self.province, 0)
+
+        actual_percentage = {
+            province["province"]: province["province__count"] * 100 / len(all_provinces)
+            for province in all_provinces.annotate(models.Count("province"))
```
rudigiesler: I feel like this would run the expensive Count per province twice, also not sure if `len(all_provinces)` will do to count every time.. I'm wondering if that's fine, or if we should only do the count once.

`assertNumQueries` is a good way to check how many and what queries using unit tests.

Also, getting the values for all provinces, and then only returning a single province, seems a bit wasteful. Maybe something like:
```
all_total = all_provinces.count()
province_total = all_provinces.filter(province=self.province).count()
province_percentage = province_total * 100 / all_total
```

eventstore/models.py
erikh360: I just added this because the randomisation relies on it being there for all 3 studies. Some of our tests don't specify a province then it fails.

rudigiesler: Should it not maybe then be `or`, so that if they're under 18 or they don't have a province, then it's skipped?

erikh360: yeah 'or' would be correct

rudigiesler (approved):

PR #413, opened 2021-11-05
Add send media template endpoint

erikh360:


eventstore/tasks.py
```diff
@@ -532,3 +532,49 @@ def process_ada_assessment_notification(
...
+    time_limit=15,
+)
+def send_whatsapp_media_template(wa_id, template_name, media_url):
+    headers = {
+        "Authorization": "Bearer {}".format(settings.TURN_TOKEN),
+        "Accept": "application/vnd.v1+json",
+        "Content-Type": "application/json",
+    }
+
+    filename = media_url.split("/")[-1]
```
rudigiesler, line 552: Could maybe use a combination of `urlparse(media_url).path` and `os.path.basename` instead, to handle all of the edge cases (eg. query strings), but this is good too since we'll be controlling the URL.

rudigiesler (approved):
:+1: looks good. Are we using this from RapidPro? I'm wondering if we couldn't just use a webhook block instead?

Also, if we're going to be sending a lot of the same media item, it might be useful to upload the media and cache the media ID, so that we can reuse that media ID for each message, instead of creating a new media ID for every message.

erikh360:
Thanks @rudigiesler this is for sending the privacy policy so it will be for USSD too. 

I guess I can make a flow and rather call that from USSD, probably better for history etc. 🤔

PR #414, opened 2021-11-10
Handle whatsapp failure events too

erikh360:


eventstore/test_whatsapp_actions.py
```diff
@@ -216,6 +216,36 @@ def test_handle_edd_message(self):
...
+        event.timestamp = timezone.now() + timedelta(days=2)
+
+        DeliveryFailure.objects.create(number_of_failures=4, contact_id="27820001001")
+
+        with patch("eventstore.tasks.rapidpro") as p:
+            handle_event(event)
+
+        p.create_flow_start.assert_called_once_with(
+            extra={
+                "optout_reason": "whatsapp_failure",
```
rudigiesler, line 238: Just noting that we must remember to update the flow as well, to handle this new reason (it needs to be mapped to a number to submit to openhim)

rudigiesler (approved):

eventstore/test_whatsapp_actions.py
erikh360: awesome, thanks!

PR #418, opened 2021-12-13
Change ask feedback into more generic model

erikh360:
Not in prod yet so model delete shouldn't be a problem


eventstore/migrations/0051_auto_20211213_1038.py
```diff
@@ -0,0 +1,50 @@
...
+                ),
+                (
+                    "data",
+                    django.contrib.postgres.fields.jsonb.JSONField(
+                        blank=True, default=dict, null=True
+                    ),
+                ),
+            ],
+        ),
+        migrations.DeleteModel(name="AskFeedback"),
```
rudigiesler, line 49: This migrate is doing a create model + delete model. Just want to make sure that this isn't something we have data in, otherwise we're going to lose all that data instead of migrating it to the new model.

rudigiesler (approved):

eventstore/migrations/0051_auto_20211213_1038.py
erikh360: Yeah its not in prod yet so all good, thanks for reviewing!

PR #419, opened 2021-12-22
Random contacts information

Buhle79:
@erikh360 can you please review


eventstore/random_contacts.py
```diff
@@ -0,0 +1,109 @@
...
+
+@app.task(
+    autoretry_for=(RequestException, SoftTimeLimitExceeded, TembaHttpError),
+    retry_backoff=True,
+    max_retries=15,
+    acks_late=True,
+    soft_time_limit=10,
+    time_limit=15,
+)
+def get_contacts():
```
erikh360: You can move this to eventstore/tasks.py

eventstore/random_contacts.py
```diff
@@ -0,0 +1,109 @@
...
+
+        if response:
+            if response["ok"]:
+                return True
+            return False
+    except Exception as e:
+        return False
+
+
+if __name__ == "__main__":
```
erikh360: I don't think we need this

eventstore/random_contacts.py
erikh360: Lets make this name more descriptive, something like `post_random_contacts_to_slack`

eventstore/test_random_contacts.py
```diff
@@ -0,0 +1,134 @@
+from django.test import TestCase, override_settings
+from temba_client.v2 import TembaClient
+
+from eventstore import random_contacts
+import responses
+
+
+class HandleRapidProContactsTests(TestCase):
```
erikh360: We can also move this into test_tasks.py, and rename to be `WhateverTaskNameTests`

eventstore/random_contacts.py
```diff
@@ -0,0 +1,109 @@
...
+                turn_link = profile.json().get("chat").get("permalink")
+        except Exception as e:
+            return turn_link
+    return turn_link
+
+
+def send_slack_message(contact_details):
+    # Send message to slack
+    try:
+        response = requests.post(
```
erikh360: We can add a if statement before this post to check if SLACK_URL and SLACK_TOKEN has a value

eventstore/random_contacts.py
```diff
@@ -0,0 +1,109 @@
...
+from django.conf import settings
+from ndoh_hub.celery import app
+from celery.exceptions import SoftTimeLimitExceeded
+from requests.exceptions import RequestException
+from temba_client.exceptions import TembaHttpError
+import datetime
+from ndoh_hub.utils import rapidpro, get_today
+
+
+if settings.EXTERNAL_REGISTRATIONS_V2:
```
erikh360: Do we need this if statement?

eventstore/random_contacts.py
```diff
@@ -0,0 +1,109 @@
...
+    if settings.TURN_URL and settings.TURN_TOKEN:
+        turn_api = settings.TURN_URL + "/v1/contacts/{}/messages"
+        turn_header = {
+            "Authorization": "Bearer {}".format(settings.TURN_TOKEN),
+            "Accept": "application/vnd.v1+json",
+        }
+
+    if settings.RAPIDPRO_URL and settings.RAPIDPRO_TOKEN:
```
erikh360: Don't think we need to check if settings.RAPIDPRO_TOKEN exists, we're not using it

eventstore/random_contacts.py
```diff
@@ -0,0 +1,109 @@
...
+                                "Turn Link": profile_link,
+                            }
+
+                            contact_details.append(links)
+
+        if contact_details:
+            sent = send_slack_message(contact_details)
+            return {"success": sent, "results": contact_details}
+        return {"success": False, "results": contact_details}
+    except Exception as e:
```
erikh360: We don't want to trap the exceptions here, we want the task to be retried on the configured exceptions and we want to be notified by sentry for any other exceptions. The sentry part happens automatically for uncaught exceptions.

I see there are a few `try..except` blocks, not going to comment on them all but the same thing goes

erikh360 (commented):
Nice work 🍰 

added a few comments

eventstore/test_random_contacts.py
Buhle79: Do you mean i should move the entire test to test_tasks.py?

erikh360: Yes, we usually have all our tasks in `tasks.py` and all our task related tests in `test_tasks.py`.

eventstore/tasks.py
```diff
@@ -532,3 +538,75 @@ def process_ada_assessment_notification(
...
+        if contact_number:
+            turn_url = settings.TURN_URL + "/v1/contacts/{}/messages".format(
+                contact_number
+            )
+
+            profile = requests.get(url=turn_url, headers=turn_header)
+
+            if profile:
+                # Get turn message link
+                return profile.json().get("chat").get("permalink")
```
erikh360: This will fail if the `chat` key doesn't exist, you can return {} as the default in the first `.get()`

eventstore/tasks.py
```diff
@@ -532,3 +538,75 @@ def process_ada_assessment_notification(
...
+)
+def post_random_contacts_to_slack_channel():
+    # Get 10 random contacts to post to slack channel
+    if settings.RAPIDPRO_URL and settings.RAPIDPRO_TOKEN:
+        rapidpro_url = urljoin(settings.RAPIDPRO_URL, "/contact/read/{}/")
+
+        contact_details = []
+
+        while len(contact_details) < settings.RANDOM_CONTACT_LIMIT:
+            profile_link, contact_uuid = get_random_contact()
```
erikh360: Can we rename `profile_link` to `turn_profile_link` please?

eventstore/tasks.py
```diff
@@ -532,3 +538,75 @@ def process_ada_assessment_notification(
...
+
+            rapidpro_link = rapidpro_url.format(contact_uuid)
+            links = {"Rapid Pro Link: ": rapidpro_link, "Turn Link": profile_link}
+
+            contact_details.append(links)
+
+        if contact_details:
+            sent = send_slack_message("test-mon", str(contact_details))
```
erikh360: I'm not sure this `str()` of a dict will look great in slack, we'll need to do some formatting.

and lets make the channel configurable with an environment variable. we can also use this to disable the task for QA etc

erikh360 (commented):
few more comments

eventstore/tasks.py
```diff
@@ -532,3 +537,81 @@ def process_ada_assessment_notification(
...
+    acks_late=True,
+    soft_time_limit=10,
+    time_limit=15,
+)
+def post_random_contacts_to_slack_channel():
+    # Get 10 random contacts to post to slack channel
+    if settings.RAPIDPRO_URL and settings.RAPIDPRO_TOKEN:
+        rapidpro_url = urljoin(settings.RAPIDPRO_URL, "/contact/read/{}/")
+
+        contact_details = {}
```
erikh360: Lets rather make this a list:
```
contacts = []
...
contacts.append(f"{contact_number} - {rapidpro_link} {turn_profile_link}")
...
send_slack_message(settings.SLACK_CHANNEL, "\n".join(contacts))
```

ndoh_hub/settings.py
```diff
@@ -236,7 +236,7 @@
 CELERY_TASK_DEFAULT_QUEUE = "ndoh_hub"
 CELERY_TASK_QUEUES = (Queue("ndoh_hub", Exchange("ndoh_hub"), routing_key="ndoh_hub"),)
 
-CELERY_TASK_ALWAYS_EAGER = False
+CELERY_TASK_ALWAYS_EAGER = True
```
erikh360: Don't think we need this change

ndoh_hub/utils.py
```diff
@@ -215,3 +218,30 @@ def authenticate(self, request):
...
+    if settings.SLACK_URL and settings.SLACK_TOKEN:
+        response = requests.post(
+            urljoin(settings.SLACK_URL, "/api/chat.postMessage"),
+            {"token": settings.SLACK_TOKEN, "channel": channel, "text": text},
+        ).json()
+
+        if response:
+            if response["ok"]:
+                return True
+            return False
```
erikh360: We can probably remove the first `return False`

eventstore/test_tasks.py
```diff
@@ -703,3 +704,188 @@ def test_valid(self):
...
+                    "unread_count": 0,
+                    "uuid": "68h4b3-6a4e-4962-8ed-c572c36fdd",
+                }
+            },
+        )
+
+        responses.add(
+            responses.POST,
+            "http://slack.com/api/chat.postMessage",
+            json={
```
erikh360: This json is what the http request will return, not what it expects.

You should rather check the request after running the task to check the request body and url.

Example: https://github.com/praekeltfoundation/ndoh-hub/blob/develop/eventstore/test_tasks.py#L97-L114

erikh360 (commented):
Just three small comments then we're good

erikh360 (approved):
🍰 🎉  Merge! lets try it out in QA

PR #422, opened 2022-01-11
slack text message link tags

Buhle79:
Hide text message links and set cron for script


ndoh_hub/settings.py
```diff
@@ -257,11 +257,22 @@
     "HANDLE_EXPIRED_HELPDESK_CONTACTS_HOUR", "3"
 )
 
+RANDOM_CONTACTS_HOUR = env.str("RANDOM_CONTACTS_HOUR", "9")
```
erikh360: Let's change this to 2 please, so that it's available as soon as we open slack and it doesn't interfere with anything else.

erikh360 (approved):
Awesome, just the one small comment but looks good

PR #423, opened 2022-01-18
library upgrade and python version upgrade

Buhle79:
Celery upgrade


eventstore/test_hcs_tasks.py
```diff
@@ -12,7 +12,7 @@ class UpdateTurnContactTaskTest(TestCase):
     def test_update_turn_contact_task(self):
         responses.add(
             responses.PATCH,
-            f"https://turn/v1/contacts/27820001001/profile",
+            f"{'https://turn/v1/contacts/27820001001/profile'}",
```
erikh360: for these we can just remove the `f` since we're not replacing anything

eventstore/test_hcs_tasks.py
```diff
@@ -22,6 +22,6 @@ def test_update_turn_contact_task(self):
         [patch] = responses.calls
 
         self.assertEqual(
-            patch.request.url, f"https://turn/v1/contacts/27820001001/profile"
+            patch.request.url, f"{'https://turn/v1/contacts/27820001001/profile'}"
```
erikh360: same as above, so just `"https://turn/v1/contacts/27820001001/profile"` no f or {}

eventstore/test_tasks.py
```diff
@@ -91,7 +91,7 @@ def test_conversation_expired(self):
         )
 
         responses.add(
-            responses.POST, f"http://turn/v1/chats/27820001001/archive", json={}
+            responses.POST, f"{'http://turn/v1/chats/27820001001/archive'}", json={}
```
erikh360: same

eventstore/test_tasks.py
```diff
@@ -111,7 +111,7 @@ def test_conversation_expired(self):
             json.loads(turn_archive.request.body),
             {
                 "before": "ABGGJ4NjeFMfAgo-sCqKaSQU4UzP",
-                "reason": f"Auto archived after 11 days",
+                "reason": f"{'Auto archived after 11 days'}",
```
erikh360: same

ndoh_hub/settings.py
```diff
@@ -139,7 +139,7 @@
 
 USE_L10N = True
 
-USE_TZ = True
+# USE_TZ = True
```
erikh360: Not sure if we need this change?

ndoh_hub/testsettings.py
```diff
@@ -1,4 +1,4 @@
-from ndoh_hub.settings import *  # flake8: noqa
+from ndoh_hub.settings import REST_FRAMEWORK  # flake8: noqa
```
erikh360: undo

setup.py
```diff
@@ -16,10 +16,11 @@
         "Markdown==3.1.1",
         "dj-database-url==0.5.0",
         "django-environ==0.4.5",
-        "psycopg2==2.7.5",
+        # "psycopg2==2.7.5",
```
erikh360: remove commented out line

erikh360 (commented):
looks good, some small comments

eventstore/serializers.py
```diff
@@ -397,3 +424,28 @@ class Code(serializers.Serializer):
 
 class ForgetContactSerializer(serializers.Serializer):
     contact_id = serializers.UUIDField(required=True)
+
+
+class RegistrationSerializer(serializers.ModelSerializer):
```
erikh360: Should we not leave this one in the registration app? its not used in eventstore

eventstore/admin.py
```diff
@@ -233,3 +236,27 @@ def get_readonly_fields(self, request, obj=None):
         if obj is None:
             return ()
         return super().get_readonly_fields(request, obj)
+
+
+class RegistrationAdmin(admin.ModelAdmin):
```
erikh360: I don't think this should be here?

eventstore/tasks.py
```diff
@@ -111,6 +160,58 @@ def get_rapidpro_contact_by_msisdn(context, field):
     return context
 
 
+@app.task()
+def remove_personally_identifiable_fields(registration_id):
```
erikh360: Don't need this

eventstore/tasks.py
```diff
@@ -52,6 +96,11 @@ def get_utc_now():
     return datetime.now(tz=pytz.utc)
 
 
+is_client = IdentityStoreApiClient(
```
erikh360: Don't need this

scripts/migrate_to_rapidpro/fix_auto_baby_switch_hub.py
```diff
@@ -19,7 +19,7 @@ async def process_babyswitch(session, row, target, jembi_writer):
     contact_id = row["contact_id"]
     timestamp = row["timestamp"]
 
-    url = urllib_parse.urljoin(HUB_URL, f"/api/v2/babyswitches/")
+    url = urllib_parse.urljoin(HUB_URL, f"{'/api/v2/babyswitches/'}")
```
erikh360: Don't need the F and {''}

erikh360 (commented):
Looking good, just a few things I think we don't need

setup.py
```diff
@@ -44,7 +44,7 @@
         "Framework :: Django",
         "License :: OSI Approved :: BSD License",
         "Programming Language :: Python",
-        "Programming Language :: Python :: 3.6",
+        "Programming Language :: Python :: 3.9.9",
```
rudigiesler, line 47: I don't think there's a 3.9.9 classifier, only 3.9: https://pypi.org/classifiers/

eventstore/tasks.py
```diff
@@ -618,3 +661,129 @@ def get_random_contact():
...
+        pass
+    for message_type in ("image", "document", "audio", "video", "voice", "sticker"):
+        try:
+            return message[message_type].get("caption", "<{}>".format(message_type))
+        except KeyError:
+            pass
+    try:
+        assert "contacts" in message
+        return "<contacts>"
+    except AssertionError:
```
rudigiesler: This is a silly pattern that I used a while ago, might be a good opportunity to change it.

Using try/except is a good python pattern, but using `assert` in anything other than unit tests has issues, most notably that they can be disabled: https://docs.python.org/3/reference/simple_stmts.html#the-assert-statement

A better pattern here would be just an if statement.

rudigiesler (commented):
Just two small comments, but overall looks great!

rudigiesler (approved):

erikh360 (approved):

PR #430, opened 2022-02-08
changed turn url

Buhle79:


eventstore/tasks.py
```diff
@@ -626,7 +626,7 @@ def get_turn_profile_link(contact_number):
         }
 
         if contact_number:
-            turn_url = settings.TURN_URL + "/v1/contacts/{}/messages".format(
+            turn_url = settings.TURN_URL + "v1/contacts/{}/messages".format(
```
erikh360: Let rather use the `urljoin` package to build this URL, then we'll know it will always work, doesn't matter what TURN_URL is configured as.

erikh360: Example: https://github.com/praekeltfoundation/ndoh-hub/blob/develop/eventstore/tasks.py#L71

eventstore/tasks.py
```diff
@@ -626,7 +626,7 @@ def get_turn_profile_link(contact_number):
         }
 
         if contact_number:
-            turn_url = settings.TURN_URL + "/v1/contacts/{}/messages".format(
+            turn_url = urljoin(settings.TURN_URL, "v1/contacts/{}/messages").format(
```
erikh360: Sorry for adding this in a separate review but please change this to an f string:
`turn_url = urljoin(settings.TURN_URL, f"v1/contacts/{contact_number}/messages")`

erikh360 (approved):

PR #436, opened 2022-03-01
strata randomization api

Buhle79:


mqr/views.py
```diff
@@ -3,11 +3,50 @@
 from rest_framework import generics, permissions
 from rest_framework.response import Response
 
+from .models import MqrStrata
+from .serializers import MqrStrataSerializer
+
+arms = ["ARM", "RCM", "BCM", "RCM_BCM", "RCM_SMS"]
```
erikh360: Lets change this variable name to uppercase and maybe a bit more descriptive like STUDY_ARMS

mqr/views.py
```diff
@@ -3,11 +3,50 @@
...
+
 
 class RandomArmView(generics.GenericAPIView):
     permission_classes = (permissions.IsAuthenticated,)
 
     def get(self, request):
         # TODO: change this to the strata sequential assignment
```
erikh360: You can replace this view, same endpoint should be fine

mqr/views.py
```diff
@@ -3,11 +3,50 @@
...
+    permission_classes = (permissions.IsAuthenticated,)
+    serializer_class = MqrStrataSerializer
+
+    def post(self, request):
+        serializer = MqrStrataSerializer(data=request.data)
+        serializer.is_valid(raise_exception=True)
+
+        province = serializer.validated_data.get("province")
+        weeks_pregnant = serializer.validated_data.get("weeks_pregnant")
+        age = serializer.validated_data.get("age")
```
erikh360: Remember these two have buckets:
Weeks pregnant (3 ranges: 16-20 weeks, 21-25 weeks, 26-30 weeks)
Age (2 ranges: 18-30, 31+)

So before creating the record put them in buckets

mqr/views.py
```diff
@@ -3,11 +3,50 @@
...
+        )
+
+        if created:
+            random.shuffle(arms)
+            random_arms = arms
+            strata.order = ",".join(arms)
+        else:
+            random_arms = strata.order.split(",")
+
+        arm = arms[strata.next_index]
```
erikh360: This should use the random_arms list

mqr/models.py
```diff
@@ -0,0 +1,9 @@
+from django.db import models
+
+
+class MqrStrata(models.Model):
+    province = models.CharField(max_length=25, null=False, blank=False)
+    weeks_pregnant = models.CharField(max_length=12, null=False, blank=False)
+    age = models.IntegerField(null=False, blank=False)
```
erikh360: Make this a charfield too since it needs to be a bucket/range, see other comment for more details

mqr/serializers.py
```diff
@@ -0,0 +1,9 @@
+from rest_framework import serializers
+
+
+class MqrStrataSerializer(serializers.Serializer):
+    province = serializers.CharField(required=True)
+    weeks_pregnant = serializers.CharField(required=True)
+    age = serializers.IntegerField(required=True)
```
erikh360: Same as above

mqr/tests/test_randomizations.py
```diff
@@ -0,0 +1,65 @@
+from django.contrib.auth import get_user_model
```
erikh360: These tests can move to test_views.py

erikh360 (commented):
Looking good but have a few comments, think you also need to generate the migration file

mqr/serializers.py
```diff
@@ -1,6 +1,14 @@
 from rest_framework import serializers
 
 
+class MqrStrataSerializer(serializers.Serializer):
+    province = serializers.CharField(required=True)
+    weeks_pregnant_bucket = serializers.CharField(required=True)
+    age_bucket = serializers.CharField(required=True)
+    next_index = serializers.IntegerField(required=False)
+    order = serializers.CharField(required=False)
```
erikh360: These two won't be sent by Rapidpro to the endpoint so we don't need them here.

mqr/tests/test_models.py
```diff
@@ -0,0 +1,46 @@
+from django.test import TestCase
```
erikh360: I'm not sure if these tests are needed, they seems to be testing the django `get_or_create` function, it's mostly safe to assume those are tested by them

mqr/tests/test_views.py
```diff
@@ -121,3 +122,72 @@ def test_faq_message(self, mock_get_message_details):
...
+            self.url,
+            data={
+                "province": "EC",
+                "weeks_pregnant_bucket": "21-25",
+                "age_bucket": "31+",
+            },
+            format="json",
+        )
+
+        self.assertEqual(type(response.data), str)
```
erikh360: We can get the strata record from the DB and do a compare here

mqr/tests/test_views.py
```diff
@@ -121,3 +122,72 @@ def test_faq_message(self, mock_get_message_details):
...
+                "weeks_pregnant_bucket": "28-30",
+                "age_bucket": "31+",
+            },
+            format="json",
+        )
+
+        splitted_arms = get_arm.order.split(",")
+
+        self.assertNotEqual(splitted_arms[0], response.data)
+        self.assertEqual(splitted_arms[1], response.data)
```
erikh360: We're hard coding the order so we can replace both these lines with: 
`self.assertEqual(response.data, "RCM_BCM")`

mqr/views.py
```diff
@@ -6,14 +6,46 @@
...
+            random.shuffle(STUDY_ARMS)
+            random_arms = STUDY_ARMS
+            strata.order = ",".join(random_arms)
+        else:
+            random_arms = strata.order.split(",")
+
+        arm = random_arms[strata.next_index]
+
+        if strata.next_index == len(random_arms):
+            strata.delete()
```
erikh360, line 43: Need a test to make sure this happens, can also create on with index=4, then hit endpoint, record should not exist anymore

mqr/views.py
```diff
@@ -6,14 +6,46 @@
...
+        else:
+            strata.next_index += 1
+            strata.save()
 
-    def get(self, request):
-        # TODO: change this to the strata sequential assignment
-        random_arm = random.choice(["ARM", "RCM", "BCM", "RCM_BCM", "RCM_SMS"])
-        return Response({"random_arm": random_arm})
+        return Response(arm)
```
erikh360: Can we change this back to return a dict `{"random_arm": random_arm}`, its just easier to handle in rapidpro

erikh360 (commented):
Looking good, few more comments.

mqr/tests/test_views.py
```diff
@@ -53,7 +54,7 @@ def test_next_message(self, mock_get_message_details):
                 "message": "Test Message 1",
                 "is_template": False,
                 "next_send_date": next_date,
-                "tag": "BCM_week_PRE19",
+                "tag": "BCM_week_PRE18",
```
erikh360: I'll need to make a more permanent fix for this

mqr/tests/test_views.py
```diff
@@ -121,3 +122,120 @@ def test_faq_message(self, mock_get_message_details):
...
+
+        MqrStrata.objects.create(
+            province="EC",
+            weeks_pregnant_bucket="21-25",
+            age_bucket="31+",
+            next_index=3,
+            order="RCM_SMS,RCM_BCM,RCM,ARM,BCM",
+        )
+
+        response = MqrStrata.objects.get(
```
erikh360: I would suggest calling the api endpoint here, then asserting on the status code and response

mqr/tests/test_views.py
```diff
@@ -121,3 +122,120 @@ def test_faq_message(self, mock_get_message_details):
...
+                "weeks_pregnant_bucket": "28-30",
+                "age_bucket": "31+",
+            },
+            format="json",
+        )
+
+        splitted_arms = get_arm.order.split(",")
+
+        self.assertNotEqual(splitted_arms[0], "RCM_BCM")
+        self.assertEqual(splitted_arms[1], "RCM_BCM")
```
erikh360: I would remove this and just add an assert on the response body, to make sure its the correct arm

mqr/tests/test_views.py
```diff
@@ -121,3 +122,120 @@ def test_faq_message(self, mock_get_message_details):
...
+        # This api call will delete the existing arm
+        self.client.post(
+            self.url,
+            data={
+                "province": "FS",
+                "weeks_pregnant_bucket": "26-30",
+                "age_bucket": "31+",
+            },
+            format="json",
+        )
```
erikh360: I think after this post we can also just assert the status code + response body
then assert that the strata record is deleted, don't need the second api call

mqr/tests/test_views.py
erikh360: Maybe change this test so that there is no strata before calling the endpoint
check status code
query strata and compare body to the strata that is found

erikh360 (commented):
few more things

mqr/tests/test_views.py
```diff
@@ -129,3 +130,105 @@ def test_faq_message(self, mock_get_message_details):
...
+
+        strata_arm = MqrStrata.objects.get(
+            province="EC", weeks_pregnant_bucket="21-25", age_bucket="31+"
+        )
+
+        self.assertContains(response, "random_arm")
```
erikh360: I would rather check the index and that the response contains the first item in the random list

erikh360: self.assertContains(response.data["random_arm"], strata_arm.order.split(",")[0])
self.assertEqual(strata_arm.next_index, 1)

Only need these two

Buhle79: Done

erikh360 (approved):

PR #440, opened 2022-03-08
Strata random arms

Buhle79:
Changed api to pass mom age and facility code

### DIFF
--- a/mqr/utils.py
+++ b/mqr/utils.py
@@ -1,4 +1,4 @@
-from datetime import datetime, timedelta
+from datetime import timedelta
 from urllib.parse import urljoin
 
 import requests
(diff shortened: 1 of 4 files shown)


mqr/utils.py
```diff
@@ -70,3 +71,49 @@ def get_next_send_date():
...
+        clinic_code = None
+
+    return clinic_code
+
+
+def get_weeks_pregnant(edd):
+    full_term_weeks = 40
+    estimated_date = edd
+
+    if isinstance(edd, str):
```
erikh360: I think if we make the serializer field a date we wont need this

mqr/utils.py
```diff
@@ -70,3 +71,49 @@ def get_next_send_date():
...
+    elif 21 <= weeks_pregnant <= 25:
+        return "21-25"
+    elif 26 <= weeks_pregnant <= 30:
+        return "26-30"
+    else:
+        return None
+
+
+def get_age_bucket(age):
+    mom_age = int(age)
```
erikh360: I think we can also make the serializer field for this an int, then we won't need this

erikh360 (commented):
Nice!, just a few small comments

mqr/utils.py
Buhle79: Done

mqr/utils.py
Buhle79: Done

erikh360 (approved):

PR #444, opened 2022-03-10
Hcs study b

bearnard:
* Adds a new model for calculating randomization targets using `source` (WhatsApp/USSD)
* Adds a view that when posted to will create a new randomization arm or update and return it if it exists already


eventstore/test_views.py
```diff
@@ -2389,6 +2390,93 @@ def test_get_list(self):
...
+        response = self.client.post(
+            self.url,
+            {
+                "msisdn": "+27820001001",
+                "source": "WhatsApp",
+                "province": "ZA-EC",
+            },
+        )
+        self.assertEqual(response.status_code, status.HTTP_200_OK)
+        print(response.json())
```
erikh360: stray print

ndoh_hub/settings.py
```diff
@@ -356,6 +356,11 @@
 HCS_STUDY_A_ACTIVE = env.bool("HCS_STUDY_A_ACTIVE", False)
 HCS_STUDY_A_WHITELIST = env.list("HCS_STUDY_A_WHITELIST", default=["*"])
 HCS_STUDY_A_CREATED_BY = env.str("HCS_STUDY_A_CREATED_BY", "whatsapp_healthcheck")
+HCS_STUDY_B_ACTIVE = env.bool("HCS_STUDY_B_ACTIVE", True)
+HCS_STUDY_B_WHITELIST = env.list("HCS_STUDY_B_WHITELIST", default=["*"])
+HCS_STUDY_B_CREATED_BY = env.list(
+    "HCS_STUDY_B_CREATED_BY", default=["whatsapp_healthcheck", "ussd_healthcheck"]
```
erikh360: This should probably default to higherhealth

erikh360 (approved):
Two small comments but looks good.

PR #449, opened 2022-03-15
Fix BCM_RCM arm

erikh360:


bearnard:
👍

Buhle79 (commented):
+1

erikh360:
Reverting, this was actually correct.

PR #454, opened 2022-03-16
Remove find part for main menu faqs

erikh360:
We don't want to add the FIND part to the bottom of the suggested FAQs for the main menu because its already there.


DevChima:
+1

Buhle79 (commented):
+1

PR #455, opened 2022-03-17
Add baseline survey model

erikh360:


mqr/models.py
```diff
@@ -7,3 +8,169 @@ class MqrStrata(models.Model):
     age_bucket = models.CharField(max_length=12, null=False, blank=False)
     next_index = models.IntegerField(default=0, null=False, blank=False)
     order = models.CharField(max_length=50, null=False, blank=False)
+
+
+class BaselineSurveyResult(models.Model):
+    class YesNoSkip:
+        YES = "yes"
```
erikh360, line 16: @Buhle79 where your choice values are 1, 2, 3 you can replace it with yes, no, skip. This just adds an extra level of validation and makes it easier for DS

Buhle79 (approved):

PR #461, opened 2022-03-23
Add GET and PATCH for baseline Survey results

erikh360:


mqr/serializers.py
```diff
@@ -9,6 +10,12 @@ class Meta:
         fields = "__all__"
         read_only_fields = ("id", "created_by")
 
+    def update(self, instance, validated_data):
+        if validated_data.get("airtime_sent"):
+            validated_data["airtime_sent_at"] = timezone.now()
```
rudigiesler, line 15: Do we need a check on current field value? eg. If you do two updates, both with `airtime_sent` set, then the first timestamp will be overwritten. Not sure if that's what we want, or if we're okay with that, or if that's not a scenario that we need to worry about.

rudigiesler (commented):
Just one small comment, if it's not an issue, then 👍

mqr/serializers.py
erikh360: Hmm good point, the flag is so that it only happen once but maybe I need some more error handling here...

Buhle79 (approved):

PR #463, opened 2022-03-24
Feature AH-141 presentation layer

DevChima:
Only a trimmed messages is presented to the user in rapidpro and the only point of interaction with the user is a single Rapidpro input node. 

Fields sent from Rapidpro
{
        "contact_uuid": "",     ----->
        "choices": "",     -----> (we need this for validation in PL)
        "value": "",    -----> (This is the user's input)
        "cardType": "",       -----> (ADA API needs this)
        "step": "",    ----> (PL and ADA API needs this)
        "optionId": "",    ----->(PL and ADA API needs this)
        "path": "",     -----> (PL needs this for the next request)
        "title": "",      -----> (DS required)
        "explanation":"",      ------> (So if there's a validation error in the PL and we send them back to RP we retain the content)
    }

Presentation layer builds the request to ADA based on the cardType as below:

TEXT cardType
request = {"step": step}

INPUT cardType
request = {"step": step, "value": userInput}

CHOICE cardType
request = {"step": step, optionId": optionId}

* Presentation Layer (PL)


Buhle79 (approved):

ada/utils.py
```diff
@@ -1,8 +1,268 @@
...
+# Go back to previous question
+def previous_question(body, path):
+    head = {
+        "x-ada-clientId": settings.X_ADA_CLIENTID,
+        "x-ada-userId": settings.X_ADA_USERID,
+        "Accept-Language": "en-GB",
+        "Accept": "application/json",
+    }
+
+    path = f"{settings.ADA_START_ASSESSMENT_URL}{path}"
```
erikh360: Please use urljoin

ada/utils.py
```diff
@@ -1,8 +1,268 @@
...
+def get_report(data):
+    head = {
+        "x-ada-clientId": settings.X_ADA_CLIENTID,
+        "x-ada-userId": settings.X_ADA_USERID,
+        "Accept-Language": "en-GB",
+        "Accept": "application/json",
+    }
+    path = data["_links"]["report"]["href"]
+    payload = {}
+    path = f"{settings.ADA_START_ASSESSMENT_URL}{path}"
```
erikh360: Please use urljoin

ada/utils.py
```diff
@@ -1,8 +1,268 @@
...
+def abort_assessment(body):
+    head = {
+        "x-ada-clientId": settings.X_ADA_CLIENTID,
+        "x-ada-userId": settings.X_ADA_USERID,
+        "Accept-Language": "en-GB",
+        "Accept": "application/json",
+    }
+
+    path = body["path"]
+    path = f"{settings.ADA_START_ASSESSMENT_URL}{path}"
```
erikh360: Please use urljoin

ada/utils.py
```diff
@@ -1,8 +1,268 @@
...
+    url = reverse("ada-reports")
+    response = requests.post(url, data=json.dumps(data), headers=head)
+    return response
+
+
+# This returns the report of the assessment
+def get_report(data):
+    head = {
+        "x-ada-clientId": settings.X_ADA_CLIENTID,
+        "x-ada-userId": settings.X_ADA_USERID,
```
erikh360: I thought one of these two need to be the msisdn or something specific to the user?

I also see we repeat this a lot, maybe we can define it in one place and grab it from there?

erikh360 (commented):
Just a few comments but looks good overall, didn't go into too much depth because of the size.

ada/utils.py
DevChima: That should be the assessmentid but it is returned as a response in ADA's payload. Something like this:

"next": {
                    "method": "POST",
                    "href": (
                        "/assessments/"
                        "654f856d-c602-4347-8713-8f8196d66be3/dialog/next"
                    ),
                },
                
                
 However, I realized the ID wasn't being saved so the latest pr saves it.

DevChima: That should be the assessmentid but it is returned as a response in ADA's payload. Something like this:

"next": {
                    "method": "POST",
                    "href": (
                        "/assessments/"
                        "654f856d-c602-4347-8713-8f8196d66be3/dialog/next"
                    ),
                },
                
                
 However, I realized the ID wasn't being saved so the latest pr saves it.

ada/utils.py
```diff
@@ -253,16 +217,20 @@ def previous_question(body, path):
...
+    head = header()
+    path = body["path"]
+    path = urljoin(settings.ADA_START_ASSESSMENT_URL, path)
+    path = path.replace("dialog/next", "/abort")
+    payload = {}
+    response = requests.put(path, json=payload, headers=head).json()
+    return response
+
+
+def header():
```
erikh360: maybe `get_header()` rather

erikh360 (approved):

PR #466, opened 2022-04-04
fixed typo for marital status

Buhle79:


erikh360 (approved):
Nice!

We might need to delete the survey records from the QA db before deploying to get this migration to work

PR #474, opened 2022-04-21
Add Choice model, uppercase keywords, renamed model, text corrections

DevChima:
-  Changes model name - Previous had a trailing 's' so it 'ss' in django.
- Adds 'choice' model and paramater - This is supposed to hold a list of options and save the content of the choice the user made so it's easier to read and analyze from a DS perspective (DS requirement)
- Did uppercase comparisons for keywords so user input is case insensitive
- Improved a couple of expressions


DevChima:
Pushing changes to QA to test.

PR #490, opened 2022-05-24
Roadblocks for ada assessment

DevChima:
Handle roadblock cardType sent from ADA
Also fix MQR test causing the build to fail.


ada/models.py
```diff
@@ -62,6 +62,7 @@ class AdaSelfAssessment(models.Model):
     created_at = models.DateTimeField(auto_now_add=True)
     choice = models.TextField(null=True, blank=True)
     pdf_media_id = models.CharField(max_length=255, null=True, blank=True)
+    roadblock = models.TextField(null=True, blank=True)
```
erikh360: Should this be a text field? looks like we only save ROADBLOCK in there?

erikh360 (commented):
I'll fix the MQR test properly at a later stage

ada/models.py
DevChima: Yes @erikh360 . Initially wasn't sure what we would save there. I'll change to CharField

Buhle79:
+1

PR #497, opened 2022-06-22
Style changes

DevChima:
Additional style changes so we have something like this but with a title: 

Have you been diagnosed with high blood pressure?

1. Yes
2. No
3. I'm not sure

Reply BACK to go to the previous question
Reply EXPLAIN to see what this means.


Buhle79 (approved):
+1

PR #498, opened 2022-06-22
Handle ADA timeout state

DevChima:
ADA now has a session timeout of +2hrs. 
If timeout, add TIMEOUT cardType to the payload and send to Rapidpro.
Rapidpro splits on cardType.


Buhle79 (approved):
+1

PR #502, opened 2022-07-05
Feature ada beta updates

DevChima:
Remove CTA at the end of assessment
Bold and Dash number formating for options list


Buhle79:
+1

PR #503, opened 2022-07-14
ADA Beta changes and MQR test fix

DevChima:
*EXIT* CTA if first step of assessment
Show only custom titles
Fix MQR test


KaitCrawford (approved):
👍

PR #506, opened 2022-07-18
Add explanation as an option

DevChima:
If an explanation exists for the question, add as choice in options list.


fritzbrand:
👍

PR #512, opened 2022-11-01
Mc aaq app

fritzbrand:


aaq/tasks.py
```diff
@@ -0,0 +1,35 @@
...
+        "feedback_secret_key": secret_key,
+        "inbound_id": inbound_id,
+        "feedback": {"feedback_type": feedback_type},
+    }
+    if "faq_id" in kwargs:
+        data["feedback"]["faq_id"] = kwargs["faq_id"]
+    if "page" in kwargs:
+        data["feedback"]["page_number"] = kwargs["page"]
+
+    url = f"{settings.AAQ_CORE_API_URL}/inbound/feedback"
```
erikh360: You can use `urllib.parse.urljoin` here instead, that way it will still work if `AAQ_CORE_API_URL` is configured with trialing slash or not.

aaq/tasks.py
```diff
@@ -0,0 +1,35 @@
...
+    headers = {
+        "Authorization": settings.AAQ_CORE_INBOUND_CHECK_TOKEN,
```
erikh360, line 32: Does the API expect the Auth token like this? no `Token` or `Bearer` in front of the token.

aaq/tasks.py
```diff
@@ -0,0 +1,35 @@
...
+from celery.exceptions import SoftTimeLimitExceeded
+from django.conf import settings
+from requests.exceptions import RequestException
+from temba_client.exceptions import TembaHttpError
+
+from ndoh_hub.celery import app
+
+
+@app.task(
+    autoretry_for=(RequestException, SoftTimeLimitExceeded, TembaHttpError),
```
erikh360: `TembaHttpError` is a Rapidpro API client error, shouldn't be possible here

ndoh_hub/testsettings.py
```diff
@@ -26,3 +26,8 @@
 
 MQR_CONTENTREPO_URL = "http://contentrepo"
 MQR_SEND_AIRTIME_FLOW_ID = "mqr-send-airtime-flow-uuid"
+
+# AAQ_HUB_APP_API_URL = env.str("AAQ_HUB_APP_API_URL", None)
```
erikh360: Remove the commented out ones to prevent confusion later on.

ndoh_hub/settings.py
```diff
@@ -387,3 +388,11 @@
 MQR_CONTENTREPO_URL = env.str("MQR_CONTENTREPO_URL", None)
 MQR_STUDY_START_DATE = env.str("MQR_STUDY_START_DATE", "2022-05-09")
 MQR_SEND_AIRTIME_FLOW_ID = env.str("MQR_SEND_AIRTIME_FLOW_ID", None)
+
+# AAQ-Beta
+AAQ_HUB_APP_API_URL = env.str("AAQ_HUB_APP_API_URL", None)
```
erikh360, line 393: Are we using all of these?

aaq/views.py
```diff
@@ -0,0 +1,116 @@
...
+        )
+    if "page_number" in json_data["feedback"]:
+        page_number = json_data["feedback"]["page_number"]
+        send_feedback_task.apply_async(
+            args=[feedback_secret_key, inbound_id, feedback_type],
+            kwargs={
+                "page": page_number,
+            },
+            countdown=5,
+        )
```
erikh360: Lets chat in slack about this part.

aaq/views.py
```diff
@@ -0,0 +1,116 @@
...
+@api_view(("POST",))
+@renderer_classes((JSONRenderer,))
+def get_first_page(request, *args, **kwargs):
+
+    serializer = InboundCheckSerializer(data=request.data)
+    serializer.is_valid(raise_exception=True)
+
+    question = serializer.validated_data["question"]
+
+    url = f"{settings.AAQ_CORE_API_URL}/inbound/check"
```
erikh360: Use urljoin as mentioned before

aaq/views.py
```diff
@@ -0,0 +1,116 @@
...
+
+    response = requests.request("POST", url, json=payload, headers=headers)
+    top_responses = response.json()["top_responses"]
+
+    json_msg = {}
+    body_content = {}
+    message_titles = []
+
+    counter = 0
+    for id, title, content in top_responses:
```
erikh360: You can use `enumerate` here to avoid having to manually increment `counter`

erikh360 (commented):
- I think we can removef the empty file `admin.py`.
- Also  remove the print statements if they aren't needed.
- Looks like we are missing tests for the task and two of the views

Looking good and I think it's a technically sound solution, just a couple of minor comments.

aaq/tasks.py
fritzbrand: I have the Token or Bearer part included as part of the secret value

ndoh_hub/settings.py
fritzbrand: I believe so yes

aaq/tasks.py
fritzbrand: Removed

aaq/tasks.py
fritzbrand: Done

ndoh_hub/testsettings.py
fritzbrand: Done

aaq/tasks.py
erikh360: Oh ok, we usually add it here, but either way works.

Having it in config makes it easier to change if they decide to change what they accept, so you might be on to something.

aaq/views.py
fritzbrand: Done

aaq/views.py
fritzbrand: Done

aaq/views.py
fritzbrand: Have implemented

erikh360 (approved):
Looks good

I see we are still missing tests for the add feedback views and there is some commented out lines.

Happy for that to be fixed in a separate PR

fritzbrand:
@erikh360 sorry I forgot about your comment saying those tests can be done in a separate PR, so added them to this one.  Can you check and let me know how I can improve those tests?

aaq/tests/test_views.py
```diff
@@ -0,0 +1,239 @@
...
+
+        payload = json.dumps(data)
+        response = self.client.put(
+            self.url, data=payload, content_type="application/json"
+        )
+        assert response.status_code == 202
+
+    @responses.activate
+    def test_page_feedback_view(self):
+        """Test that we can submit feedback on an FAQ"""
```
erikh360, line 164: This comment should probably say page

erikh360 (approved):
Those look good.

The only thing I would suggest is testing the validation by sending invalid things and making sure the API responds accordingly.

PR #525, opened 2023-01-12
Mqr 276 move randomization

Hlamallama:
separate the validation and randomization of the arms

### DIFF
--- a/mqr/views.py
+++ b/mqr/views.py
@@ -77,6 +77,7 @@ class RandomStrataArmValidationView(generics.GenericAPIView):
             )
 
 class RandomStrataArmView(generics.GenericAPIView):
+
     def post(self, request):
         """
         Randomization of the ARMs.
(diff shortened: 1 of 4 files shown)


mqr/views.py
```diff
@@ -64,6 +64,39 @@ def post(self, request):
...
         if clinic_code and weeks_pregnant_bucket and age_bucket:
+            return Response(
+                {
+                    "Valid": True,
+                }
+            )
+
+        return Response(
+                {
+                    "Valid": False,
```
erikh360: I think we should still include the reason here. It saves time if you can see the result in the rapidpro contact history

mqr/views.py
```diff
@@ -39,7 +39,7 @@
 STUDY_ARMS = ["ARM", "RCM", "RCM_BCM", "RCM_SMS"]
 
 
-class RandomStrataArmView(generics.GenericAPIView):
+class RandomStrataArmValidationView(generics.GenericAPIView):
```
erikh360: We can probably remove the random part of the class name here.

erikh360 (commented):
Only the two comments for now but looks good.

I think we can remove the reference to random in all the validation things, but confusing

erikh360 (approved):

PR #530, opened 2023-02-28
Handle timeouts

DevChima:
Contacts have around 2hrs to complete an assessment after which ADA returns a 410.
At the moment this throws a HTTPError in Sentry and then slack #momconnectmonitoring

These alerts are not necessary since there's nothing we can do about it. 
This pr should stop the noise in slack since 410 HTTPError's will no longer fire.


bearnard:
👍

DevChima:
Thanks @bearnard

PR #531, opened 2023-03-23
Handle new recipient id

erikh360:


eventstore/test_views.py
```diff
@@ -1468,6 +1467,86 @@ def test_successful_events_request(self, mock_handle_event):
...
+        self.assertEqual(event.created_by, user.username)
+        self.assertFalse(event.fallback_channel)
+        self.assertEqual(event.data, {"random": "data"})
+
+        mock_handle_event.assert_not_called()
+
+    @mock.patch("eventstore.views.handle_event")
+    def test_successful_events_request_missing_recipient_id(self, mock_handle_event):
+        """
+        Should create a new Event object in the database
```
rudigiesler: This docstring doesn't look correct

eventstore/views.py
```diff
@@ -166,7 +166,14 @@ def create(self, request):
 
             for statuses in request.data.get("statuses", []):
                 message_id = statuses.pop("id")
-                recipient_id = statuses.pop("recipient_id")
+
+                if "message" in statuses:
+                    recipient_id = statuses["message"].pop("recipient_id")
+                    if statuses["message"] == {}:
+                        statuses.pop("message")
```
rudigiesler, line 173: Not sure if this is something we'll see in practice, but what about if we have something like `{"message": {}, "recipient_id": "abc"}`, then we won't extract the recipient ID correctly here, but will have an error instead. Maybe a try/catch instead of `if "message" in statuses` might be better?

rudigiesler (commented):
Looks good! Just two small comments

Buhle79 (commented):
+1

eventstore/views.py
erikh360: I think the validation on the serialiser will trap this

rudigiesler: Wouldn't the validation pass because `recipient_id` exists? It just checks if either `message.recipient_id` OR `recipient_id` exists, and in the example `{"message": {}, "recipient_id": "abc"}`, I think that's true? Either way, I don't think this is a thing we'll see in production, happy to let it be.

rudigiesler (approved):

PR #532, opened 2023-03-24
Upgrade to Django 4

erikh360:


eventstore/tasks.py
```diff
@@ -731,6 +731,9 @@ def get_text_or_caption_from_turn_message(message: dict) -> str:
     if message["type"] is None:
         return "<unknown>"
 
+    if message["type"] == "button":
+        return "<button>"
```
rudigiesler: We could probably return the content of the button instead of `<button>`, since that's more in-line with how it's displayed on the whatsapp UI

changes/models.py
```diff
@@ -3,7 +3,7 @@
 from django.contrib.auth.models import User
 from django.contrib.postgres.fields import JSONField
 from django.db import models
-from django.utils.encoding import python_2_unicode_compatible
+from six import python_2_unicode_compatible
```
rudigiesler: We can probably just remove these instead, since in python 3 this does nothing, and we're definitely not running this under python 2 anymore

rudigiesler (commented):
Overall looks good! I'm not sure if you've tested this, but maybe we should test that changing to the new JSON field, even though it makes migration files, the migrations don't make any actual changes to the database (otherwise it will be a long and slow migration). I think running `sqlmigrate` against a test postgresql database will show you the exact sql that will be sent to the database.

erikh360:
@rudigiesler thanks for the review. Haven't done much testing but I did check all the migrations and the json field migrations are all noops.

rudigiesler (approved):

PR #542, opened 2023-04-04
Add index to event.recipient_id concurrently

erikh360:
The table has over 215mil rows so we need to do this concurrently.

Followed this example:
https://www.agiliq.com/blog/2021/07/django-concurrent-indexes-without-down-time/


Buhle79 (commented):
+1

PR #545, opened 2023-04-13
Delay forget_contact task

erikh360:
We need to give all the messages and events time to get into the hub before remove the pii.


fritzbrand:
👍

PR #546, opened 2023-04-14
Update whatsapp contact check

erikh360:
This contacts endpoint is being deprecated and always returns valid, they're also sending out warnings about using the endpoint.


DevChima:
+1

PR #549, opened 2023-05-02
Rename assessment report

DevChima:
Rename file name to "{record_id}_report" from "report" so it's easier to sort.

### DIFF
--- a/ada/utils.py
+++ b/ada/utils.py
@@ -2,6 +2,7 @@ from __future__ import absolute_import, division
 
 import json
 import posixpath
+import re
 import tempfile
 import urllib.parse
 from urllib.parse import urlencode, urljoin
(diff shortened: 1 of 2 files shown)


ada/utils.py
```diff
@@ -340,6 +340,7 @@ def upload_edc_media(report, study_id, record_id, field_id, token):
     study_data_point = "study-data-point"
     field_id = field_id
     path = posixpath.join(study_id, record, record_id, study_data_point, field_id)
+    report_name = record_id + "_" + "report"
```
erikh360: This will not have the `.json` extension and record_id might have something not fit for a file name, so we should probably clean it.

erikh360 (commented):
Just the one comment and we should probably get a unit test here

ada/utils.py
```diff
@@ -340,6 +341,8 @@ def upload_edc_media(report, study_id, record_id, field_id, token):
     study_data_point = "study-data-point"
     field_id = field_id
     path = posixpath.join(study_id, record, record_id, study_data_point, field_id)
+    report_name = record_id + "_" + "report.json"
```
erikh360, line 344: This could be: `f"{record_id}_report.json"`

erikh360 (approved):
Nice, one small comment but happy for that to be addressed later on

PR #550, opened 2023-05-03
send template to whatsapp

Hlamallama:


ndoh_hub/test_utils.py
```diff
@@ -21,3 +27,52 @@ def test_msisdn_to_whatsapp_id(self):
...
+        )
+
+        response = send_whatsapp_template_message(
+            msisdn, namespace, template_name, parameters
+        )
+
+        self.assertEqual(response, "Whatsapp")
+
+
+class TestSendWhatsappTemplateMessageFalse(TestCase):
```
erikh360: This could be in the same class as above and maybe be renamed to `TestSendWhatsappTemplateMessage`

erikh360 (commented):
Looks good just the one small comment

ndoh_hub/test_utils.py
Hlamallama: @erikh360 I have updated it

erikh360 (approved):

PR #552, opened 2023-05-08
Momza 2292 sms registration

Hlamallama:


ndoh_hub/urls.py
```diff
@@ -97,4 +99,9 @@
     path(
         "metrics", internal_only(django_prometheus.ExportToDjangoView), name="metrics"
     ),
+    re_path(
+        r"^api/v1/send-whatsapp-template",
```
erikh360: We usually don't have '-' in paths

ndoh_hub/serializers.py
```diff
@@ -0,0 +1,12 @@
+from rest_framework import serializers
+
+
+class WhatsappTemplateMessageSerializer(serializers.Serializer):
+    class ParametersField(serializers.Serializer):
```
erikh360, line 5: Please add some tests for invalid requests

erikh360 (commented):
Just the two small comments

What is the test_decorators file for and why was it renamed?

Hlamallama:
> Just the two small comments
> 
> What is the test_decorators file for and why was it renamed?

Moved it and test_utils into a tests folder

erikh360 (approved):

PR #568, opened 2023-06-19
Add privacy policy PDF

erikh360:
Adding it here so that we can add the link to a sms for Non WA ussd registrations


DevChima:
+!

PR #579, opened 2023-08-30
Add delete historical management command

erikh360:


rudigiesler (approved):
Looks good! The only thing I can think of checking is does the database get locked when doing the delete? For especially the first deletes, that could mean that we can't do any inserts for a long period of time. I think that's why RapidPro gets all the IDs of messages to delete, and then deletes them in chunks: https://github.com/nyaruka/rp-archiver/blob/main/archives/messages.go#L170

erikh360:
Thanks @rudigiesler I didn't even think about table locking, according to the docs it looks like delete uses a RowExclusiveLock so it won't block other operations from completing.

It might still be useful to do it in chunks thought but I think I'll look at that in a separate PR.

PR #590, opened 2023-11-23
Add openhim queue model and signal

erikh360:


eventstore/models.py
```diff
@@ -1094,3 +1094,39 @@ def clean(self):
...
+        CHW_REGISTRATION = "CHWRegistration"
+        PUBLIC_REGISTRATION = "PublicRegistration"
+        CHANNEL_SWITCH = "ChannelSwitch"
+        OPTOUT = "OptOut"
+        choices = (
+            (PREBIRTH_REGISTRATION, "PrebirthRegistration"),
+            (CHW_REGISTRATION, "CHWRegistration"),
+            (PUBLIC_REGISTRATION, "PublicRegistration"),
+            (CHANNEL_SWITCH, "ChannelSwitch"),
+            (OPTOUT, "OptOut"),
```
rudigiesler, line 1111: Not a big thing, but I think we usually use the snake case for these values, eg. `prebirth_registration`, for the value that gets stored in the database, and then title case for the values that get shown in the UI, eg. `Prebirth Registration`

eventstore/models.py
```diff
@@ -1094,3 +1094,39 @@ def clean(self):
...
+        PROCESSING = 1
+        COMPLETE = 2
+        ERROR = 3
+
+        choices = (
+            (PENDING, "Pending"),
+            (PROCESSING, "Processing"),
+            (COMPLETE, "Complete"),
+            (ERROR, "Error"),
+        )
```
rudigiesler: Are we on a new enough version of django that we can use `models.IntegerChoices` instead? https://docs.djangoproject.com/en/4.2/ref/models/fields/#enumeration-types

eventstore/signals.py
```diff
@@ -0,0 +1,13 @@
+from django.db.models.signals import post_save
+from django.dispatch import receiver
+
+from eventstore.models import OpenHIMQueue, PrebirthRegistration
+
+
+@receiver(post_save, sender=PrebirthRegistration)
+def create_user_profile(sender, instance, created, **kwargs):
```
rudigiesler: Should this function name be something like `queue_prebirth_registration`?

eventstore/models.py
```diff
@@ -1094,3 +1094,39 @@ def clean(self):
...
+
+    id = models.UUIDField(primary_key=True, default=uuid.uuid4, editable=False)
+    object_id = models.UUIDField()
+    object_type = models.CharField(max_length=30, choices=ObjectType.choices)
+    status = models.PositiveSmallIntegerField(
+        choices=Status.choices, default=Status.PENDING
+    )
```
rudigiesler, line 1125: Do we also need to store a copy of the details from the object? In case they do a forget delete before this gets forwarded? Or in that case do we not want to forward the object to them?

eventstore/models.py
erikh360: Thanks for the review, yes that makes sense, I initially wanted to get the models dymanically so tried to keep them the same as the class name

eventstore/signals.py
erikh360: Yes

eventstore/models.py
erikh360: I'll check

eventstore/models.py
erikh360: Hmm I didn't think about that use case, I would assume we don't send it then, since they requested we delete all their data. We can't leave it in a queue then

erikh360: I've noted the assumption on the ticket

rudigiesler (approved):

PR #613, opened 2024-08-01
Add AAQ v2 faq search endpoint

Hlamallama:


aaq/views.py
```diff
@@ -116,3 +117,37 @@ def check_urgency(request, *args, **kwargs):
...
+@renderer_classes((JSONRenderer,))
+def search(request, *args, **kwargs):
+    serializer = SearchSerializer(data=request.data)
+    serializer.is_valid(raise_exception=True)
+    query_text = serializer.validated_data["query_text"]
+    generate_llm_response = serializer.validated_data["generate_llm_response"]
+    query_metadata = serializer.validated_data["query_metadata"]
+    url = urllib.parse.urljoin(settings.AAQ_V2_API_URL, "search")
+    payload = {
+        "query_text": f"{query_text}",
```
erikh360: These probably don't have to be f-strings

aaq/views.py
```diff
@@ -116,3 +117,37 @@ def check_urgency(request, *args, **kwargs):
...
+        "query_metadata": f"{query_metadata}",
+    }
+    headers = {
+        "Authorization": settings.AAQ_V2_AUTH,
+        "Content-Type": "application/json",
+    }
+
+    response = requests.request("POST", url, json=payload, headers=headers)
+
+    json_msg = {
```
erikh360, line 168: This looks like its just recreating the dict exactly how we got it from AAQ and returning it?

Shouldn't we build a message here like AAQ v1?

aaq/views.py
```diff
@@ -116,3 +117,37 @@ def check_urgency(request, *args, **kwargs):
...
+        "query_id": response.json()["query_id"],
+        "llm_response": response.json()["llm_response"],
+        "search_results": response.json()["search_results"],
+        "feedback_secret_key": response.json()["feedback_secret_key"],
+        "debug_info": response.json()["debug_info"],
+        "state": response.json()["state"],
+    }
+
+    return_data = json_msg
```
erikh360: This extra variable isn't really required

erikh360 (commented):
Left a few comments.

The commits and PR could be better named, ie "Add AAQ v2 faq search endpoint"

aaq/views.py
DevChima: @Hlamallama Have have you seen this feedback?

Hlamallama: @DevChima yes, i've implemeted the feedback from Eric

DevChima (approved):

PR #616, opened 2024-08-06
aaq urgency check endpoint V2

Hlamallama:
Search endpoint to call aaq seatch and urgency detect

### DIFF
--- a/aaq/tests/test_views.py
+++ b/aaq/tests/test_views.py
@@ -535,7 +535,7 @@ class CheckUrgencyV2ViewTests(APITestCase):
     @responses.activate
     def test_urgency_check_urgent(self):
         """
-        Test that we can get an urgency score of 1.0
+        Test that we can get is urgent True
         """
         user = get_user_model().objects.create_user("test")
         self.client.force_authenticate(user)
@@ -569,7 +569,7 @@ class CheckUrgencyV2ViewTests(APITestCase):
     @responses.activate
     def test_urgency_check_not_urgent(self):
         """
-        Test that we can get an urgency score of 0.0
+        Test that we can get is urgent False
         """
         user = get_user_model().objects.create_user("test")
         self.client.force_authenticate(user)
--- a/aaq/tests/test_views.py
+++ b/aaq/tests/test_views.py
@@ -554,7 +554,17 @@ class CheckUrgencyV2ViewTests(APITestCase):
         )
 
         assert response.status_code == 200
-        assert response.json() == {"is_urgent": True}
+        assert response.json() == {
+            "details": {
+                "0": {"distance": 0.1, "urgency_rule": "Blurry vision and dizziness"},
+                "1": {"distance": 0.2, "urgency_rule": "Nausea that lasts for 3 days"},
+            },
+            "is_urgent": True,
+            "matched_rules": [
+                "Blurry vision and dizziness",
+                "Nausea that lasts for 3 days",
+            ],
+        }
 
     @responses.activate
     def test_urgency_check_not_urgent(self):
@@ -578,7 +588,14 @@ class CheckUrgencyV2ViewTests(APITestCase):
         )
 
         assert response.status_code == 200
-        assert response.json() == {"is_urgent": False}
+        assert response.json() == {
+            "details": {
+                "0": {"distance": 0.1, "urgency_rule": "Baby okay"},
+                "1": {"distance": 0.2, "urgency_rule": "Baby healthy"},
+            },
+            "is_urgent": False,
+            "matched_rules": ["Baby okay", "Baby healthy"],
+        }
 
     @responses.activate
     def test_urgency_check_invalid(self):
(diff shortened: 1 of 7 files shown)


aaq/views.py
```diff
@@ -222,3 +223,29 @@ def search(request, *args, **kwargs):
...
+    serializer.is_valid(raise_exception=True)
+    message_text = serializer.validated_data["message_text"]
+    url = urllib.parse.urljoin(settings.AAQ_V2_API_URL, "/urgency-check-v2")
+    headers = {
+        "Authorization": settings.AAQ_V2_API_URL,
+        "Content-Type": "application/json",
+    }
+
+    response = requests.request("POST", url, json=message_text, headers=headers)
+    is_urgent = response.json()["is_urgent"]
```
erikh360: I think we should also return true/false to the rp flow, then there is no guessing what 0/1 means.

erikh360 (commented):
Looking at the return body we could just return that straight to the flow.

aaq/views.py
```diff
@@ -222,3 +223,21 @@ def search(request, *args, **kwargs):
...
+    serializer.is_valid(raise_exception=True)
+    message_text = serializer.validated_data["message_text"]
+    url = urllib.parse.urljoin(settings.AAQ_V2_API_URL, "/urgency-check-v2")
+    headers = {
+        "Authorization": settings.AAQ_V2_API_URL,
+        "Content-Type": "application/json",
+    }
+
+    response = requests.request("POST", url, json=message_text, headers=headers)
+    is_urgent = {"is_urgent": response.json()["is_urgent"]}
```
erikh360: We could just change this to:
`return Response(response.json(), status=status.HTTP_200_OK)`

aaq/tests/test_views.py
```diff
@@ -527,3 +527,71 @@ def test_response_feedback_invalid_feedback_text_view(self):
...
         }
+
+
+class CheckUrgencyV2ViewTests(APITestCase):
+    url = reverse("aaq-check-urgency-v2")
+
+    @responses.activate
+    def test_urgency_check_urgent(self):
+        """
+        Test that we can get an urgency score of 1.0
```
erikh360: Update these docstrings to match the true/false

aaq/tests/test_views.py
```diff
@@ -527,3 +527,88 @@ def test_response_feedback_invalid_feedback_text_view(self):
...
+        """
+        user = get_user_model().objects.create_user("test")
+        self.client.force_authenticate(user)
+        fakeAaqUdV2Api = FakeAaqUdV2Api()
+        responses.add_callback(
```
erikh360: Can you validate the request payload that is sent to the AAQ endpoint please?

Hlamallama: i have this test test_urgency_check_invalid to check if the payload is valid

erikh360: That is the payload to our endpoint, I'm talking about the payload e send to the aaq endpoint

aaq/tests/test_views.py
```diff
@@ -527,3 +529,90 @@ def test_response_feedback_invalid_feedback_text_view(self):
...
+            responses.POST,
+            "http://aaq_v2/urgency-check-v2",
+            callback=fakeAaqUdV2Api.post_urgency_detect_return_true,
+            content_type="application/json",
+        )
+
+        payload = json.dumps({"message_text": "I am pregnant and out of breath"})
```
erikh360: If you remove the .dumps here, you can remove the 2 .loads() calls on line 571

aaq/views.py
```diff
@@ -173,21 +173,27 @@ def response_feedback(request, *args, **kwargs):
...
     }
+    search_url = urllib.parse.urljoin(settings.AAQ_V2_API_URL, "/search")
+
+    if "generate_llm_response" in serializer.validated_data:
+        search_payload["generate_llm_response"] = serializer.validated_data[
+            "generate_llm_response"
+        ]
+    if "query_metadata" in serializer.validated_data:
+        search_payload["query_metadata"] = serializer.validated_data["query_metadata"]
+
```
erikh360: Can we give these defaults in the serializer and always include them?

erikh360 (commented):
Just the one comment for now.

Can we maybe move some of this functionality into a utils.py file to separate things a bit. Lets chat in slack about this

aaq/utils.py
```diff
@@ -0,0 +1,80 @@
...
+        "body": body_content,
+        "feedback_secret_key": feedback_secret_key,
+        "query_id": query_id,
+    }
+
+    urgency_check_payload = {
+        "message_text": query_text,
+    }
+
+    check_urgency_response = check_urgency_v2(urgency_check_payload)
```
erikh360: Do you need to send a dict here? You build a dict inside `check_urgency_v2` too?

If the tests are passing it means somewhere we're not checking something

aaq/utils.py
```diff
@@ -0,0 +1,80 @@
...
+
+    json_msg.update(check_urgency_response.data)
```
erikh360: Can we use .json() instead of .data? Probably applies to a few places

erikh360 (commented):
Starting to look good, just a few comments

erikh360 (approved):

PR #632, opened 2024-10-16
Hub cleanup - Remove mqr

DevChima:


DevChima:
This PR removes mqr from Hub.

PR #638, opened 2024-10-17
add test coverage to PR comment using py-cov-action

Hlamallama:


erikh360 (approved):
Nice! but probably remove the utils test part before merging

PR #639, opened 2024-10-22
Repo consistency

fritzbrand:
This PR is for the repo consistency work done on this repo.

The main goal is getting all repos to use poetry for package management, and ruff for linting.


Dockerfile
```diff
@@ -1,10 +1,12 @@
 FROM ghcr.io/praekeltfoundation/docker-django-bootstrap-nw:py3.9-bullseye
 
-COPY setup.py /app
-RUN pip install --no-cache-dir -e .
+RUN pip install poetry==1.8.3
```
jerith: Perhaps remove the version pin and move it to just above the `RUN poetry config` line?

eventstore/tests/test_views.py
```diff
@@ -1390,19 +1395,21 @@ def test_successful_outbound_messages_request(self, mock_handle_outbound):
...
+                messages.data,
+                {
+                    "text": {"body": "your-text-message-content"},
+                    "render_mentions": True,
+                    "preview_url": True,
+                    "random": "data",
+                    "recipient_type": "individual",
+                },
+            ),
+        )
```
erikh360: I think we can just remove the trailing comma that was probably added as a mistake. They don't need to be tuples

eventstore/tests/test_whatsapp_actions.py
```diff
@@ -201,7 +201,7 @@ def test_handle_edd_label_disabled(self):
 
         with patch("eventstore.whatsapp_actions.handle_edd_message") as handle:
             handle_inbound(message)
-            handle.assert_not_called
+            _var = handle.assert_not_called
```
erikh360: I think this was also a mistake, it should have been `handle.assert_not_called()`

erikh360 (commented):
Nice, just the two small changes

eventstore/tests/test_views.py
```diff
@@ -1283,38 +1283,43 @@ def test_successful_inbound_messages_request(self, mock_handle_inbound):
...
-                    "name": "location-name",
+        (
+            self.assertEqual(
+                messages.timestamp,
+                datetime.datetime(2018, 2, 15, 11, 38, 20, tzinfo=UTC),
+            ),
+        )
+        (self.assertEqual(messages.id, "9e12d04c-af25-40b6-aa4f-57c72e8e3f91"),)
+        (self.assertEqual(messages.type, "image"),)
+        (self.assertEqual(messages.message_direction, Message.INBOUND),)
```
jerith: This looks like `ruff` has helpfully turned a bunch of accidental trailing commas into explicit tuples. Perhaps we should remove the commas instead?

eventstore/tests/test_whatsapp_actions.py
jerith: This should be a method call, I think:
```suggestion
            handle.assert_not_called()
```

scripts/growgreat_extract/msisdn_from_details.py
```diff
@@ -4,7 +4,7 @@
 
 def get_msisdn(details):
     last = None
-    for addr, addr_dets in details.get("addresses", {}).get("msisdn", {}).items():
+    for addr, _addr_dets in details.get("addresses", {}).get("msisdn", {}).items():
```
jerith: If we're not using the value, do we need to iterate over `.items()`?

fritzbrand: I have updated it as follows, and the tests still pass.  Is this what you meant?

```
for addr in details.get("addresses", {}).get("msisdn", {}):
        if details.get("default") is True:
            return addr
        if details.get("optedout") is not True:
            last = addr
    return last
```

eventstore/tests/test_whatsapp_actions.py
fritzbrand: Updated

eventstore/tests/test_views.py
fritzbrand: Fixed

Dockerfile
fritzbrand: 👍

eventstore/tests/test_whatsapp_actions.py
fritzbrand: 👍

eventstore/tests/test_views.py
fritzbrand: 👍

scripts/growgreat_extract/msisdn_from_details.py
jerith: Yup, thanks. 🙂

eventstore/tests/test_views.py
jerith: I think there are more of these -- mostly multiline ones. Any line that starts `(self.assert...` or the multiline equivalent with the `(` on the line above can be de-tupled.

erikh360 (approved):

erikh360 (approved):

PR #643, opened 2024-10-22
update gibberish response dict

Hlamallama:


aaq/utils.py
```diff
@@ -35,8 +35,8 @@ def search(query_text, generate_llm_response, query_metadata):
 
     response = requests.request("POST", url, json=payload, headers=headers)
 
-    if "error" in response.json():
-        error_detail = response.json()["error"].get("detail", "")
+    if "detail" in response.json():
```
erikh360: Are we sure detail won't be present for any other requests?

Hlamallama: a successful search response doesn't doesn't have the detail key

erikh360: Cool, but it might be confusing later on so what about something like:
`if response.status_code == 400 "detail" in response.json():`

Then theres no doubt that we are handling an error,

You'll need to make sure you mock the status code the same as the AAQ service returns it.

aaq/utils.py
```diff
@@ -35,13 +35,13 @@ def search(query_text, generate_llm_response, query_metadata):
 
     response = requests.request("POST", url, json=payload, headers=headers)
 
-    if "error" in response.json():
-        error_detail = response.json()["error"].get("detail", "")
+    if response.status_code == 400 and "detail" in response.json():
```
erikh360: I meant we should use the `status.HTTP_400_BAD_REQUEST` here, if we return 400 to rapidpro it would think the request failed

erikh360: `if response.status_code == status.HTTP_400_BAD_REQUEST and "detail" in response.json():`

Is what I meant

erikh360 (approved):

PR #648, opened 2025-01-09
Scripts to migrate contacts from Rapidpro to Turn

erikh360:


scripts/migrate_to_turn/update_turn_contacts.py
```diff
@@ -0,0 +1,49 @@
...
+async def update_turn_contact_details(session, row, target):
+    wa_id = row.pop("wa_id")
+
+    url = urljoin(TURN_URL, f"/v1/contacts/{wa_id}/profile")
+    headers = {
+        "Authorization": f"Bearer {os.environ['TURN_TOKEN']}",
+        "content-type": "application/json",
+        "Accept": "application/vnd.v1+json",
+    }
+    await request(session, url, "PATCH", headers, row, target)
```
DevChima, line 25: Looks good only one question. Is it necesary to have error handling for these API  and async requests.

DevChima (approved):

scripts/migrate_to_turn/update_turn_contacts.py
erikh360: That request function outputs the results and when running the script we pipe it all to a results file.

We then use `jq` to check for any errors and run the retry script that does the same, we can repeat that until they all work.

PR #650, opened 2025-01-14
Add update turn contacts script with asyncio queue

erikh360:


scripts/migrate_to_turn/update_turn_contacts_queue.py
```diff
@@ -0,0 +1,58 @@
...
+
+async def worker(name, queue):
+    while True:
+        session, row, target = await queue.get()
+        await update_turn_contact_details(session, row, target)
+        queue.task_done()
+
+
+async def main(filename, target):
+    queue = asyncio.Queue()
```
rudigiesler: We should add a max size to the queue (probably the number of workers), so that it blocks while adding items to the queue, so that we don't read all the rows at once. Combined with `await queue.put(update)`, and starting the workers before we start adding the tasks.

rudigiesler (commented):
Looks good! Just a comment on how we can limit the size of the queue, so that we don't read the whole file into memory at once.

rudigiesler (approved):

PR #662, opened 2025-07-02
Increment failure count for specific error codes

DevChima:
This PR refines the logic in the handle_event function to improve the accuracy of detecting when a contact is not on WhatsApp by ensuring that we only increment the failure count for specific error codes which indicate a contact is no longer on Whatsapp. 

It utilizes an included_error_codes list containing error codes that suggest the contact is not on WhatsApp or has opted out of messages on that channel. The failure count is only incremented when we get these failed events.

A test was also added to verify that the failure count does not increment when the error code is not in the included_error_codes list.

### DIFF
--- a/eventstore/whatsapp_actions.py
+++ b/eventstore/whatsapp_actions.py
@@ -95,6 +95,12 @@ def handle_event(event):
     """
     Triggers all the actions that are required for this event
     """
+
+    included_error_codes = [
+        131026,  # Not a WhatsApp number, outdated app or unaccepted terms
+        131050,  # User has stopped receipt of marketing messages
+    ]
+
     if event.status == Event.FAILED:
         reason = OptOut.WHATSAPP_FAILURE_REASON
         if event.fallback_channel is True:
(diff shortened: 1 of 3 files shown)


eventstore/whatsapp_actions.py
```diff
@@ -95,6 +95,12 @@ def handle_event(event):
     """
     Triggers all the actions that are required for this event
     """
+
+    included_error_codes = [
+        131026,  # Not a WhatsApp number, outdated app or unaccepted terms
+        131050,  # User has stopped receipt of marketing messages
```
erikh360, line 101: What does this error mean? do you have a link to the docs please?

DevChima: @erikh360 Here: https://developers.facebook.com/docs/whatsapp/cloud-api/support/error-codes
131026:
Unable to deliver message. Reasons can include:

The recipient phone number is not a WhatsApp phone number.
Recipient has not accepted our new Terms of Service and Privacy Policy.
Recipient using an old WhatsApp version

131050:
Unable to deliver the message. This recipient has chosen to stop receiving marketing messages on WhatsApp from your business.

erikh360 (approved):

PR #665, opened 2025-10-13
Nluclassifier for feedback labelling

DevChima:
This pull request add and endpoint for processing user feedback (complaints/compliments) and classifying/labeling the message in Turn.
1. Endpoint is `/label/feedback`
2. App receives a POST request from Rapidpro with `message_id` and `inbound_message` in the body.
3. It starts the celery task `process_feedback_for_labeling` which calls our NLU service for intent classification.
4. The app then retrieves the intent and calls Turn to label the message if the intent is a `COMPLIMENT` or a `COMPLAINT`.

NB: There were some black and flake8 conflicts where black would condense ta block in a single line but it would then be flagged by flake8 for being over the limit. 
To bypass that, I added an `#fmt: off` in some places so black knows to skip those lines.

### DIFF
--- a/nluclassifier/tasks.py
+++ b/nluclassifier/tasks.py
@@ -1,0 +1,119 @@
+import logging
+from urllib.parse import urljoin
+
+import requests
+from celery.exceptions import SoftTimeLimitExceeded
+from django.conf import settings
+from requests.auth import HTTPBasicAuth
+from requests.exceptions import RequestException
+
+from ndoh_hub.celery import app
+
+logger = logging.getLogger(__name__)
+
+CELERY_TASK_OPTIONS = {
+    "autoretry_for": (RequestException, SoftTimeLimitExceeded),
+    "retry_backoff": True,
+    "max_retries": 15,
+    "acks_late": True,
+    "soft_time_limit": 10,
+    "time_limit": 15,
+}
+
+
+@app.task(**CELERY_TASK_OPTIONS)
+def process_feedback_for_labeling(message_id: str, inbound_message: str):
+    """
+    Call the NLU endpoint for feedback classification.
+    LLabel the message in Turn using the detected intent.
+    """
+
+    nlu_label = None
+
+    NLU_URL = settings.INTENT_CLASSIFIER_URL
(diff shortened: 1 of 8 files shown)


Mudiwa66:
Looks good, do you want to add a test for Turn API failures?

DevChima:
Could be useful to add. Let me see. 
Thanks @Mudiwa66

Mudiwa66 (approved):

nluclassifier/views.py
```diff
@@ -0,0 +1,51 @@
...
+            )
+            return Response(
+                {
+                    "error": (
+                        "Missing 'message_id' or "
+                        "'inbound_message' parameter in POST body."
+                    ),
+                },
+                status=status.HTTP_400_BAD_REQUEST,
+            )
```
erikh360, line 43: It's usually easier to do this validation using a serializer

nluclassifier/tasks.py
```diff
@@ -0,0 +1,119 @@
...
+            f"NLU service failed for message {message_id}. "
+            f"Retrying. Error: {e}"
+            )
+        raise
+    except Exception as e:
+        logger.error(
+            f"NLU non-retriable failure for message {message_id}: "
+            f"{e}"
+            )
+        return
```
erikh360, line 66: Won't this mean the exception doesn't go to sentry?

nluclassifier/tasks.py
```diff
@@ -0,0 +1,119 @@
...
+
+        except RequestException as e:
+
+            logger.warning(
+                f"Turn API failed to label message {message_id}. "
+                f"Retrying. Error: {e}"
+                )
+            raise
+        except Exception as e:
+            logger.error(
```
erikh360, line 111: Same here

erikh360 (commented):
The sentry one might be a problem

nluclassifier/tasks.py
DevChima: Hmm, I was thinking it will log and stil be caught by Sentry?

DevChima: Yeah, Sentry shoudl catch it first before Celery. I can test this in QA.

PR #666, opened 2025-10-21
babyloss nlu

DevChima:
This PR updates existing NLU logic by calling a new babyloss endpoint and therefore prioritizing Babyloss intent for inbound messages. 
`handle_inbound()` is updated to celery tasks when an inbound is received. 
1. `get_inbound_intent` hits the NLU endpoint /nlu/babyloss/ which returns a booloean result ({"babyloss": True/False}) as per [documentation](https://docs.google.com/document/d/17Bo7rzQtI_a4ubNnV1XSaBI3NEwABvJHvcOhsxEPX1c/edit?tab=t.blzrb09cfktk#heading=h.2fmt26cvnrf6).
2. `route_nlu_result` task evaluates the result and checks the message channel (`is_sms`)
If `babyloss: True`, it triggers the `Babyloss` label function.
If `babyloss: False` and WA, the process terminates else if SMS, we classify the intent (complaint or compliment) and label the message.

### DIFF
--- a/eventstore/tasks.py
+++ b/eventstore/tasks.py
@@ -820,13 +820,10 @@ def route_nlu_result(
     If the message is SMS, it processes feedback for labeling.
     """
     if is_babyloss_intent is True:
-        label_whatsapp_message.s("BABYLOSS", message_id).delay()
+        label_whatsapp_message.delay("BABYLOSS", message_id)
         return
 
-    if is_sms is False:
-        return
-
-    if is_sms is True:
+    elif is_sms is True:
         process_feedback_for_labeling.delay(message_id, inbound_message)
         return
 
--- a/eventstore/tasks.py
+++ b/eventstore/tasks.py
@@ -840,7 +840,6 @@ def route_nlu_result(
     time_limit=15,
 )
 def label_whatsapp_message(label, message_id):
-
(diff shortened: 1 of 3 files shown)


eventstore/tasks.py
```diff
@@ -792,14 +793,42 @@ def process_whatsapp_template_send_status():
...
+    Receives the boolean result from the babyloss NLU endpoint and routes to the next step.
+    If babyloss intent is True, it labels the message as BABYLOSS.
+    If the message is SMS, it processes feedback for labeling.
+    """
+    if is_babyloss_intent is True:
+        label_whatsapp_message.s("BABYLOSS", message_id).delay()
+        return
+
+    if is_sms is False:
+        return
```
erikh360: I don't think this if statement is doing anything

erikh360: You can probably remove all the returns and just use one if with a if else to simplify. this a little

erikh360 (commented):
Few small comments

eventstore/tasks.py
DevChima: Updated the if block. Yeah that if statement was redundant.

erikh360 (approved):

PR #684, opened 2026-03-09
Add turn fields to schedule PNC appointments

erikh360:


scripts/migrate_to_turn/process_fields.py
```diff
@@ -68,58 +78,137 @@ def get_user_dob_year(value):
...
+
+def _get_next_vaccine(birth_date, now=None):
+    now = now or datetime.now(pytz.utc)
+    now_date = now.date()
+
+    for value, unit, text in VACCINE_SCHEDULE:
+        if unit == "W":
+            due_date = birth_date + timedelta(weeks=value)
+        else:
+            due_date = _add_months(birth_date, value)
```
rudigiesler, line 126: I think the `dateutil` library can deal with this well, but happy for us to implement this as well to avoid the dependency.

rudigiesler (approved):

PR #686, opened 2026-03-17
Update batch update turn contacts script

erikh360:


scripts/migrate_to_turn/update_turn_contacts_bulk.py
```diff
@@ -1,28 +1,335 @@
...
     url = urljoin(TURN_URL, "v1/contacts")
     headers = {
         "Authorization": f"Bearer {os.environ['TURN_TOKEN']}",
         "content-type": "text/csv",
     }
-    response = requests.post(url, data=data, headers=headers)
+    with open(chunk_path, "rb") as data:
+        response = requests.post(url, data=data, headers=headers, timeout=300)
+
+    result_path = results_dir / f"{Path(chunk_path).name}.response.csv"
```
rudigiesler, line 219: It seems like Turn will return an `ERROR: ...` inline in the streamed CSV result for errors. Are we going to manually search the results for that afterwards?

rudigiesler (approved):
Looks good! The splitting of CSV files seems too complex for me, but I can't think of a better way unless we can relax some contraints (can we guarantee no newlines in CSV values? Then we can just deal with lines instead of decoding CSVs. Are we expecting headers or lines >1megabyte?, etc)

scripts/migrate_to_turn/update_turn_contacts_bulk.py
erikh360: Yeah, I'll probably add a script to analyze results and get fixable/retry able errors and take it from there

erikh360:
@rudigiesler I'd rather keep to more complex but safe way of doing things, we don't know what we are going to run into with contact names etc. We control creating the csvs but there is always something. Then we'll have to cater for every little thing.

On the size, with the tests all files result in 950KB, so it includes headers.

